PR #2, opened 2019-02-13
Add clock message support

Shokman:


caguero:
+1

iche033:
Do we need to add the functions to `convert_builtin_interfaces.hpp`? This seems to compile fine without them. Maybe @caguero  know about this?

Shokman:
Is it going to work without at least the ign to ros? The ros to ign is unnecessary but I added it just in case... It is a bridge so I thought it would be incomplete without it.

iche033:
I was thinking more of the function declarations in `convert_builtin_interfaces.hpp`. Currently the new functions that exist in .cpp file are not added to the .hpp file, i.e.:

```
template<>
void
convert_ign_to_1(
  const ignition::msgs::Clock & ign_msg,
  rosgraph_msgs::Clock & ros1_msg);

template<>
void
convert_1_to_ign(
  const rosgraph_msgs::Clock & ros1_msg,
  ignition::msgs::Clock & ign_msg);
```

So not sure if need to add them or not

test/test_utils.h
```diff
@@ -173,6 +174,25 @@ namespace testing
...
+    _msg.clock.sec  = 1;
+    _msg.clock.nsec = 2;
+  }
+
+  /// \brief Compare a message with the populated for testing.
+  /// \param[in] _msg The message to compare.
+  void compareTestMsg(const rosgraph_msgs::Clock &_msg)
+  {
+    rosgraph_msgs::Clock expected_msg;
+    createTestMsg(expected_msg);
```
iche033: `expected_msg` is not being used in this function?

iche033:
ping @Shokman @caguero 

do we want to get this in for the release?

Shokman:
I think we should. Clock message is important for many messages. I am updating this PR

caguero:
> Do we need to add the functions to `convert_builtin_interfaces.hpp`? This seems to compile fine without them. Maybe @caguero know more about this?

Yes please. Add them to `convert_builtin_interfaces.hpp` and to `builtim_interfaces_factories.hpp`.

caguero:
+1 to merge this pull request before the next release.

test/test_utils.h
Shokman: fixed

Shokman:
I updated the Pr and addressed the comments.

caguero:
Could you also add the new declarations to `convert_builtin_interfaces.hpp` and to `builtin_interfaces_factories.hpp` please?

PR #4, opened 2019-02-20
Add Pose / Transform msg support

iche033:
Add conversion between `ign::msgs::Pose` and 
  * `geometry_msgs/Pose`
  * `geometry_msgs/PoseStamped`
  * `geometry_msgs/Transform`
  * `geometry_msgs/TransformStamped`

Unlike previous msgs types, this is a one-to-many mapping. This is possible since the ign::msgs::Pose msg captures most of the fields in the above ROS geometry msgs, with the following exception:
  * `geometry_msgs/TransformStamped`'s `child_frame_id` field. 

Initial plan was to introduce a `frame` field in ignition::msgs::Pose but later examination of the current ros1_ign_bridge implementation reveals that a similar ros msg field, `frame_id`, is being stored in an any field in `ign::msgs::Header`. So I took the same approach and store the `child_frame_id` in the ign header msg.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,16 +8,21 @@ between ROS 1 and Ignition Transport.
 The bridge is currently implemented in C++. At this point there's no support for
 service calls.Its support is limited to only the following message types:
 
-| ROS 1 type                  | Ignition Transport type      |
-|-----------------------------|:----------------------------:|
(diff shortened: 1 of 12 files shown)


caguero:
Looks good to me!

caguero:
+1

PR #5, opened 2019-03-14
Missing dependencies

j-rivero:
During the packaging of the package I noted that we have some missing dependencies in the package. I've also include a little fix to use the catkin variable for installing.

I'm assuming that we have no intention to install the tests artifacts: ign_publisher and ros1_publisher.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -7,11 +7,13 @@
 
   <buildtool_depend>catkin</buildtool_depend>
 
+  <depend>geometry_msgs</depend>
   <depend>roscpp</depend>
+  <depend>rostest</depend>
+  <depend>sensor_msgs</depend>
   <depend>std_msgs</depend>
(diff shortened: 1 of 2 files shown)


package.xml
```diff
@@ -7,11 +7,13 @@
 
   <buildtool_depend>catkin</buildtool_depend>
 
+  <depend>geometry_msgs</depend>
   <depend>roscpp</depend>
+  <depend>rostest</depend>
```
chapulina: I think this can be `<test_depend>`

caguero:
+1

package.xml
j-rivero: fixed aa25d2667f6c14f9c94f2f03b824a2d04051dee5

caguero:
@j-rivero , feel free to resolve conflicts and merge.

PR #6, opened 2019-03-14
Remove dependency on ign-common

chapulina:
This requires a new ign-msgs3 release including [this PR](https://bitbucket.org/ignitionrobotics/ign-msgs/pull-requests/94/add-pixel-format-enum/diff).

### DIFF
--- a/.travis/build
+++ b/.travis/build
@@ -8,10 +8,9 @@ export DEBIAN_FRONTEND=noninteractive
 
 # Dependencies.
 echo "deb [trusted=yes] http://packages.osrfoundation.org/gazebo/ubuntu-stable bionic main" > /etc/apt/sources.list.d/gazebo-stable.list
-echo "deb [trusted=yes] http://packages.osrfoundation.org/gazebo/ubuntu-prerelease bionic main" > /etc/apt/sources.list.d/gazebo-prerelease.list
 echo "deb [trusted=yes] http://packages.ros.org/ros/ubuntu bionic main" > /etc/apt/sources.list.d/ros-latest.list
 apt-get update -qq
-apt-get install -qq -y libignition-common3-dev \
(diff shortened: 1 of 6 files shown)


chapulina:
`ign-msgs` 3.1.0 has been released so this is now ready for review.

scpeters:
I updated the readme and travis script in 9465612d8aefc3817f3977600b27165fbd9cc22b

.travis/build
```diff
@@ -8,10 +8,9 @@ export DEBIAN_FRONTEND=noninteractive
 
 # Dependencies.
 echo "deb [trusted=yes] http://packages.osrfoundation.org/gazebo/ubuntu-stable bionic main" > /etc/apt/sources.list.d/gazebo-stable.list
-echo "deb [trusted=yes] http://packages.osrfoundation.org/gazebo/ubuntu-prerelease bionic main" > /etc/apt/sources.list.d/gazebo-prerelease.list
 echo "deb [trusted=yes] http://packages.ros.org/ros/ubuntu bionic main" > /etc/apt/sources.list.d/ros-latest.list
 apt-get update -qq
-apt-get install -qq -y libignition-common3-dev \
+apt-get install -qq -y libignition-msgs3-dev \
```
chapulina: @scpeters , I noticed that the code was relying on the msgs version pulled by transport. Do you think it would be better to add explicit calls for `find_package(ignition-msgs3 QUIET REQUIRED)` since we will make the installation explicit too?

scpeters: yeah, that would be good, since it is being used directly

chapulina: Done 78ceea90035d604a58ea76e4ca0250a552a33bf1

j-rivero: change dependency in the -release repo 2c87ee0

j-rivero:
+1 changes looks good to me

PR #8, opened 2019-05-02
Float

caguero:
Conversion between `std_msgs/Float32` and `ignition::msgs::Float`

### DIFF
--- a/test/test_utils.h
+++ b/test/test_utils.h
@@ -100,7 +100,7 @@ namespace testing
   /// \param[out] _msg The message populated.
   void createTestMsg(std_msgs::Float32 &_msg)
   {
-    _msg.data = 1;
+    _msg.data = 1.5;
   }
 
   /// \brief Compare a message with the populated for testing.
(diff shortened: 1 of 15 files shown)


test/test_utils.h
```diff
@@ -95,6 +96,23 @@ namespace testing
   /// ROS 1 test utils
   //////////////////////////////////////////////////
 
+  /// \brief Create a message used for testing.
+  /// \param[out] _msg The message populated.
+  void createTestMsg(std_msgs::Float32 &_msg)
+  {
+    _msg.data = 1;
```
scpeters: I would consider using a floating point value like `1.5` instead of `1`

test/test_utils.h
```diff
@@ -475,6 +493,23 @@ namespace testing
   /// Ignition::msgs test utils
   //////////////////////////////////////////////////
 
+  /// \brief Create a message used for testing.
+  /// \param[out] _msg The message populated.
+  void createTestMsg(ignition::msgs::Float &_msg)
+  {
+    _msg.set_data(1);
```
scpeters: consider `1.5` instead of `1`

test/test_utils.h
scpeters: 266c755e92c00674087e44d2f81afed4fa2fcc45

test/test_utils.h
scpeters: 266c755e92c00674087e44d2f81afed4fa2fcc45

PR #9, opened 2019-05-03
Joint states

caguero:
This is a conversion between `sensor_msgs/JointState` and `ignition::msgs::Model`.


scpeters:
I restarted travis builds, and they are now passing

PR #11, opened 2019-05-10
actuators: use generic field names

scpeters:
companion to [ign-msgs pr# 129](https://bitbucket.org/ignitionrobotics/ign-msgs/pull-requests/129/actuatorsproto-angle-position/diff)


scpeters:
I expect travis to fail since the ign-msgs pull request is in a branch and not in the nightly build

PR #15, opened 2019-05-25
Camera info

chapulina:
Bridge camera info messages.

camera_info

I'll update this PR with more instructions on how to test it soon. I need to fix a wrong rotation coming from ign-sensors.

### DIFF
--- a/src/convert_builtin_interfaces.cpp
+++ b/src/convert_builtin_interfaces.cpp
@@ -474,6 +474,13 @@ convert_1_to_ign(
     num_channels = 3;
     octets_per_channel = 2u;
   }
+  else if (ros1_msg.encoding == "32FC1")
+  {
+    ign_msg.set_pixel_format_type(
+      ignition::msgs::PixelFormatType::R_FLOAT32);
+    num_channels = 1;
+    octets_per_channel = 4u;
+  }
   else
   {
     ign_msg.set_pixel_format_type(
@@ -557,6 +564,13 @@ convert_ign_to_1(
     ros1_msg.encoding = "bgr16";
     num_channels = 3;
     octets_per_channel = 2u;
+  }
+  else if (ign_msg.pixel_format_type() ==
+      ignition::msgs::PixelFormatType::R_FLOAT32)
+  {
+    ros1_msg.encoding = "32FC1";
+    num_channels = 1;
+    octets_per_channel = 4u;
   }
   else
   {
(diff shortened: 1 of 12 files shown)


chapulina:
Usage example described [here](https://bitbucket.org/ignitionrobotics/ign-sensors/pull-requests/67/camera-info/diff#comment-103381905).

I tried also converting depth camera images but it's not going through correctly :cry:

src/convert_builtin_interfaces.cpp
```diff
@@ -474,6 +474,13 @@ convert_1_to_ign(
     num_channels = 3;
     octets_per_channel = 2u;
   }
+  else if (ros1_msg.encoding == "32FC1")
+  {
+    ign_msg.set_pixel_format_type(
+      ignition::msgs::PixelFormatType::R_FLOAT32);
+    num_channels = 1;
+    octets_per_channel = 4u;
+  }
```
chapulina: Currently, RViz shows a black image and RQt seems to show pixels that are closer as darker, but infinity is black:

depth_img

As reference, on Gazebo 9, RQt and RViz also show the inverse of Gazebo. Note though that the pixel format used on SDF is `R8G8B8`, like on the [kinect](https://bitbucket.org/osrf/gazebo_models/src/e6d645674e8a99cb7955ce3b9a1fbe26c58c41f2/kinect/model.sdf#lines-30), instead of `R_FLOAT32`, like on `ign-gazebo`'s demos.

ros1_depth

chapulina: > infinity is black

`gazebo_ros_pkgs` is [publishing NaN](https://github.com/ros-simulation/gazebo_ros_pkgs/blob/kinetic-devel/gazebo_plugins/src/gazebo_ros_depth_camera.cpp#L462) for values out of bounds.

`ign-sensors` is [publishing +-inf](https://bitbucket.org/ignitionrobotics/ign-sensors/src/6d6402ded8f11f86526797a2c53f085d0ccdc340/src/DepthCameraSensor.cc#lines-369) as defined by [REP 117](http://www.ros.org/reps/rep-0117.html). But when I echo the bridged ROS topic I don't see any infs.

chapulina: If I change `ign-sensors` to publish NaNs instead of +-inf, RViz shows the image. So I think RViz is not handling the infs properly. I'll ticket an issue there.

depth_nan

chapulina: If I change `ign-sensors` to publish NaNs instead of +-inf, RViz showed the image. So the problem was that RViz was not handling the infs properly. I opened a PR fixing RViz: https://github.com/ros-visualization/rviz/pull/1378

PR #16, opened 2019-05-31
Move package to subfolder, add metapackage

chapulina:
Builds on top of #15.

This is in preparation for adding a package dedicated to point clouds.


scpeters:
I changed the target branch from camera_info to master and will rebase

chapulina:
Ah nice catch, I totally forgot that was the case :bowing_woman:

PR #17, opened 2019-06-08
Point clouds for RGBD cameras

chapulina:
Needs:
- [x] #16 
- [x] https://bitbucket.org/ignitionrobotics/ign-sensors/pull-requests/70/
- [x] https://bitbucket.org/ignitionrobotics/ign-gazebo/pull-requests/356/
- [x] https://bitbucket.org/ignitionrobotics/ign-gazebo/pull-requests/351/

---

Current status: publishes `PointCloud2` without color information:

rgbd_flat


chapulina:
Colors fixed in faa07bd5e5622bfcad8386ef4cabdedcfffcbf84, but I'm not sure why I had to do BGR instead of RGB when populating the point cloud. I checked that the `ignition::rendering::Image` and the `sensor_msgs::Image` are storing RGB, but that seems to be swapped from the `PointCloud2` message. I'm still investigating, but this is usable and I think all other PRs can be merged.

chapulina:
This is ready for review.

I think the offset between the depth and RGB images is coming from `ign-rendering`. Note how the falling cone's position is slightly different in the depth and color images:

mismatched_imgs

scpeters:
I'm going to rebase this on meta_pkg

ros1_ign_point_cloud/src/point_cloud.cc
```diff
@@ -0,0 +1,347 @@
...
+      {
+        *iter_z = depth;
+      }
+
+      // Put image color data for each point
+      // \TODO(anyone) RGB image seems offset from depth image by 3~4 pixels in both directions
+      if (this->rgb_image_msg_.data.size() == _height * _width * 3)
+      {
+        // color
+        // \TODO(anyone) why do we need to do BHR instead of RGB?
```
azeey: When calling `modifier.setPointCloud2FieldsByString(2, "xyz", "rgb");` its creating the rgb field as 3 unsigned chars stacked in a float. Accessing the bytes of the float directly is probably why the colors are backward because they are using little endian byte order. The documentation [here](http://docs.ros.org/melodic/api/sensor_msgs/html/namespacesensor__msgs.html) suggests creating `PointCloud2Iterator`s for each color channel.

ros1_ign_point_cloud/src/point_cloud.cc
```diff
@@ -0,0 +1,347 @@
...
+    char** argv = NULL;
+    ros::init(argc, argv, "ignition", ros::init_options::NoSigintHandler);
+    ROS_INFO_NAMED("ros1_ign_point_cloud", "Initialized ROS");
+  }
+
+  // Sensor scoped name
+  auto scoped_name = ignition::gazebo::scopedName(this->dataPtr->entity_, _ecm, "/", false);
+
+  // ROS node
+  auto node_name = _sdf->Get<std::string>("node_name", scoped_name).first;
```
azeey: This is not a big deal, but the node name is always "ignition" right now, which I think is being set on line 119. The `NodeHandle` constructor's argument is the namespace. Do we even need node_name?

ros1_ign_point_cloud/src/point_cloud.cc
chapulina: Niiiice! Fixed in 7eede34757687e87f494d7450f7837be4142b09b

ros1_ign_point_cloud/src/point_cloud.cc
chapulina: Ohhh it's a pretty big deal, the parameter is completely wrong :sweat_smile: I'm confusing the ROSes. Fixed in 7eede34757687e87f494d7450f7837be4142b09b

chapulina:
Thanks for the review, @azeey , I think I've addressed all your comments.

azeey:
Looks like CI is failing because we added ignition packages to `package.xml`. Would doing `rosdep -r` help? Otherwise, we might have to revert those changes :slightly_frowning_face:

IeiuniumLux:
@azeey how about #18?

azeey:
I just tested with ogre vs ogre2 with a 1920x1080 rgbd camera. The weird offset thing you mentioned is much more visible in higher resolution in ogre2, but it doesn't happen in ogre.

**Ogre2**
ogre2

**Ogre**
ogre

scpeters:
CI is :green_heart:

chapulina:
Nice catch with ogre vs ogre2, @azeey . We can debug more on the ign-rendering side.

iche033:
for ogre2, do you know if it's the depth data that are at an offset or the color data?

chapulina:
@iche033 , taking ogre1 as ground truth, depth seems to be the wrong one. It seems to be a bit more zoomed in:

rgb_diff
depth_diff

iche033:
thanks that's helpful. I'm looking into it

iche033:
issue should be fixed by ign-rendering pull request 173:

https://bitbucket.org/ignitionrobotics/ign-rendering/pull-requests/173

PR #18, opened 2019-06-19
.travis/build: rosdep skip ignition keys

IeiuniumLux:
to fix the rosdep issue in #17


chapulina:
Thanks! This works for me, we just need to skip all the ign libs

scpeters:
Thanks to @jacobperron for the advice

scpeters:
I'll add the missing packages

azeey:
Nice! Didn't know about `skip-keys`.

PR #21, opened 2019-06-24
Point clouds from lidars

chapulina:
Adds:

* Flat color `PointCloud2` from GPU lidars
* Flat color `PointCloud2` from depth cameras

Refer to demos and READMEs for usage documentation.


nkoenig:
Works for me.

PR #25, opened 2019-07-01
Update releasing instructions

j-rivero:
With the use of bloom packages, I needed to change the way of releasing to use bloom and catkin.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -15,7 +15,7 @@ The steps to make a new release of the package are:
     - rosdep update
 
  1. Bloom it into a custom repository
-    -  BLOOM_RELEASE_REPO_BASE=https://github.com/osrf/ bloom-release --rosdistro melodic --track melodic ros1_ign
+    -  BLOOM_RELEASE_REPO_BASE=https://github.com/osrf/ bloom-release --rosdistro melodic --track melodic ros_ign
     -  [will fail fedora: ignore and continue]
     -  **IMPORTANT**: once bloom finishes pushing tags and branches to the repo, `control+c` the process before it sends a PR to rosdistro
 
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -15,8 +15,7 @@ The steps to make a new release of the package are:
     - rosdep update
 
  1. Bloom it into a custom repository
-    -  bloom-release --rosdistro melodic --track melodic ros_ign
-    -  [use repository: git@github.com:osrf/ros_ign-release.git]
+    -  BLOOM_RELEASE_REPO_BASE=https://github.com/osrf/ bloom-release --rosdistro melodic --track melodic ros1_ign
     -  [will fail fedora: ignore and continue]
     -  **IMPORTANT**: once bloom finishes pushing tags and branches to the repo, `control+c` the process before it sends a PR to rosdistro
 
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -2,14 +2,20 @@
 
 The package uses the latest versions of ignition libraries, at this moment only
 availables in the packages.osrfoundation.org repository. This fact makes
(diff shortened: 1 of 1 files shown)


scpeters:
Should someone else make the next release with these instructions?

j-rivero:
> Should someone else make the next release with these instructions?

good idea

nkoenig:
I'm trying this out.

RELEASING.md
```diff
@@ -2,14 +2,20 @@
...
 
- 1. Update metadata for packaging in:
-    https://bitbucket.org/osrf/ros1_ign_bridge-release/
+ 1. Follow the standard ROS release using catkin_generate_changelog and catkin_prepare_release
+    - http://wiki.ros.org/bloom/Tutorials/ReleaseCatkinPackage
 
- 1. Launch manually the builds in:
-    https://build.osrfoundation.org/job/ros1_ign_bridge-debbuilder/
+ 1. Install rosdep for ignition packages and run rosdep update
+    - sudo wget https://raw.githubusercontent.com/osrf/osrf-rosdep/master/ignition/ignition.yaml -O /etc/ros/rosdep/sources.list.d/00-ignition.list
```
nkoenig: I think this line should be:
```
sudo echo "yaml https://github.com/osrf/osrf-rosdep/raw/master/ignition/ignition.yaml" >> /etc/ros/rosdep/sources.list.d/00-ignition.list
```

j-rivero: you are right, thanks!

caguero:
Minor tweak in 2d3fe2b9edca9c2c289f273263150a1cafb5ebf3.

j-rivero:
We can remove the note about stopping bloom once we have this change https://github.com/ros-infrastructure/bloom/pull/548 released and replace it by: `--no-pullrequest` parameter

chapulina:
While working on the `0.8.0` release I learned a few things:

* The `0.7.1` release had been done from this repo's `ros_ign_release` branch, not `melodic`. I only realized this too late, so now there aren't `0.7.1` entries on `melodic`'s changelogs.
* I needed to make a couple of changes to `ros_ign-release`:
    * https://github.com/osrf/ros_ign-release/commit/cbcaab53ac9ace83fe6ca5dbe3a75d2171b24acf
    * https://github.com/osrf/ros_ign-release/commit/ad097aaa712fcfc211f08bcd38617d3c0b724abc
* I had to dig for a branch of `release-tools` which had `ros_ign_bridge-release.py.bash `. Opened a PR: https://bitbucket.org/osrf/release-tools/pull-requests/907

So I rebased this branch and added some pointers in https://github.com/osrf/ros_ign/pull/25/commits/feabb81ac61b59922effaf12eb506411d2a65aa5.

j-rivero:
While working in the initial idea of moving ros_ign1 -> ros_ign I made some modifcations in order to make the transition easier for people having ros_ign1. I assume that we no longer want that installations with ros_ign1 installed to automatically install ros_ign.
Not sure if the ros_ign1 and  ros_ign packages can be co-installated. If that is not the case, we need some metadata to avoid package conflicts.

chapulina:
> I assume that we no longer want that installations with ros_ign1 installed to automatically install ros_ign

Yes, we decided against it some weeks ago because it could break users without much notice. Instead, users will need to explicitly ask to download `ros_ign` to get it.

> Not sure if the ros_ign1 and ros_ign packages can be co-installated

I just co-installed them and checked that they don't collide. Since the package names are different, I think we're safe.

---

Sorry about the confusion, let me know if anything else isn't clear or if you have other concerns. Also let me know if I messed anything making the `0.8.0` release!

j-rivero:
> I just co-installed them and checked that they don't collide. Since the package names are different, I think we're safe.

Perfect, thanks for checking

PR #28, opened 2019-07-13
Bridge point cloud packed

nkoenig:
See: https://bitbucket.org/ignitionrobotics/ign-msgs/pull-requests/144 and https://bitbucket.org/ignitionrobotics/ign-sensors/pull-requests/78/output-a-pointcloudpacked-message/diff


chapulina:
I tried adding a demo in on branch `packed_demo` using the new message but the point cloud is not showing on RViz. The ROS message data is suspiciously short:

```
header:
  seq: 105
  stamp:
    secs: 10
    nsecs: 500000000
  frame_id: "model_with_lidar/link/gpu_lidar"
height: 1
width: 1
fields:
  -
    name: "x"
    offset: 0
    datatype: 7
    count: 1
  -
    name: "y"
    offset: 4
    datatype: 7
    count: 1
  -
    name: "z"
    offset: 8
    datatype: 7
    count: 1
is_bigendian: False
point_step: 12
row_step: 12
data: [0, 0, 128, 127, 0, 0, 128, 255, 0, 0, 192, 255]
is_dense: True
```

So is the Ignition data:

```
header {
  stamp {
    sec: 13
    nsec: 500000000
  }
  data {
    key: "frame_id"
    value: "model_with_lidar::link::gpu_lidar"
  }
}
field {
  name: "x"
  datatype: FLOAT32
  count: 1
}
field {
  name: "y"
  offset: 4
  datatype: FLOAT32
  count: 1
}
field {
  name: "z"
  offset: 8
  datatype: FLOAT32
  count: 1
}
height: 1
width: 1
point_step: 12
row_step: 12
data: "\000\000\200\177\000\000\200\377\000\000\300\377"
is_dense: true
```

I'm still looking into what could be going on. It could be a problem with my local setup.

nkoenig:
Your width and height are 1. Is that intentional?

chapulina:
Oh I hadn't noticed that detail on the message. I'm running [this example world](https://bitbucket.org/ignitionrobotics/ign-gazebo/src/0c60411df5f6b4a0c9b6f7fb9ea3b91e7f068f5b/examples/worlds/gpu_lidar_sensor.sdf#lines-217). The laser scans are going through fine. It could be an issue with my setup, I'll try your updated branches soon.

chapulina:
The GPU lidar works for me with the new branches. The depth cloud looks mirrored though. This is using the demo on #29 :

rgbd_bridge_mirror

nkoenig:
The mirror should be fixed: https://bitbucket.org/ignitionrobotics/ign-sensors/commits/3f70f5615bd01cc8a7639d80500a065e20858d5e?at=ign-sensors2

nkoenig:
Are you having trouble with the depth_image in rviz? I see all black.

chapulina:
> Are you having trouble with the depth_image in rviz? I see all black.

You need this fix, it hasn't been released yet: https://github.com/ros-visualization/rviz/pull/1378

nkoenig:
Ahh, thanks.

ros1_ign_bridge/src/convert_builtin_interfaces.cpp
```diff
@@ -904,24 +904,66 @@ template<>
...
 {
   convert_1_to_ign(ros1_msg.header, (*ign_msg.mutable_header()));
-
   std::cerr << "Unsupported conversion from [sensor_msgs::PointCloud2] to "
             << "[ignition::msgs::PointCloud]" << std::endl;
 }
 
 template<>
 void
 convert_ign_to_1(
```
chapulina: Since the messages are so similar, it shouldn't be a lot of trouble to add the reverse conversion, and some tests with some dummy data.

nkoenig:
Add other direction and tests.

PR #29, opened 2019-07-15
Packed demo

chapulina:
Builds on top of #28.

Add 2 demos using the point cloud bridge, one for GPU lidar and another for RGBD cameras,


nkoenig:
+1

This works for me.

PR #37, opened 2019-08-15
Adding debug and error statements

nkoenig:


ros1_ign_bridge/include/ros1_ign_bridge/factory.hpp
```diff
@@ -114,7 +115,7 @@ class Factory : public FactoryInterface
     const boost::shared_ptr<ros::M_string> & connection_header =
```
chapulina: I was looking at this function on `ros1_bridge` yesterday and noticed they [print more](https://github.com/ros2/ros1_bridge/blob/master/include/ros1_bridge/factory.hpp#L148-L188) debug info than we do. We could do the same here, using the type names:

caguero:
Should we add `rosconsole` as a new dependency? (`package.xml`, `CMakeLists.txt`, `README.md`)

chapulina:
> Should we add rosconsole

We're getting it indirectly through `roscpp`, but it wouldn't hurt to have the direct dependency :+1:

ros1_ign_bridge/include/ros1_ign_bridge/factory.hpp
nkoenig: 90ce9939f7c0218290f7c21fd20a0971a845cbaa

nkoenig:
Added rosconsole: 90ce9939f7c0218290f7c21fd20a0971a845cbaa

PR #38, opened 2019-08-15
Support unidirectional bridge topics

nkoenig:
Allow the use of @, [, or ] between the message types to indicate if a bridge should be bidirectional, to ROS, or to IGN.

### DIFF
--- a/ros1_ign_bridge/src/parameter_bridge.cpp
+++ b/ros1_ign_bridge/src/parameter_bridge.cpp
@@ -49,14 +49,14 @@ void usage()
 {
   ROS_ERROR_STREAM(
       "Bridge a collection of ROS1 and Ignition Transport topics.\n\n"
-      << "  parameter_bridge <topic@ROS1_type[@,<,>]Ign_type> .. "
+      << "  parameter_bridge <topic@ROS1_type(@,[,])Ign_type> .. "
       << " <topic@ROS1_type@Ign_type>\n\n"
       << "The first @ symbol delimits the topic name from the message types.\n"
       << "Following the first @ symbol is the ROS1 message type.\n"
-      << "The ROS1 message type is followed by an @, <, or > symbol where\n"
+      << "The ROS1 message type is followed by an @, [, or ] symbol where\n"
       << "    @  ==  a bidirectional bridge, \n"
-      << "    <  == a bridge from Ignition to ROS1,\n"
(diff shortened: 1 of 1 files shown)


nkoenig:
See: https://bitbucket.org/osrf/subt/pull-requests/270

AlejoAsd:
In [this comment](https://github.com/osrf/ros1_ign/pull/38#issue-307750795), you're mentioning `@`, `<` and `>` are used to indicate direction. However, the code suggests it's actually `[` and `]` for unidirectional communication.

nkoenig:
Fixed.

ros1_ign_bridge/src/parameter_bridge.cpp
```diff
@@ -33,14 +33,34 @@
...
+      << "The first @ symbol delimits the topic name from the message types.\n"
+      << "Following the first @ symbol is the ROS1 message type.\n"
+      << "The ROS1 message type is followed by an @, [, or ] symbol where\n"
+      << "    @  ==  a bidirectional bridge, \n"
+      << "    [  == a bridge from Ignition to ROS1,\n"
+      << "    ]  == a bridge from ROS1 to Ignition.\n"
+      << "Following the direction symbol is the Ignition Transport message "
+      << "type.\n"
+      << "A bidirectional bridge example:\n"
+      << "    parameter_bridge /chatter@std_msgs/String@ignition.msgs"
```
mjcarroll: Maybe add examples of unidirectional here.

nkoenig: Added examples for each.

azeey:
FYI, just noticed that zsh doesn't like the `[` character on the command line. I had to wrap the whole argument to parameter_bridge in quotes. Maybe we can mention that when the examples in the README are updated.

mjcarroll:
@azeey That's a function of `zsh`'s globbing.  If you have `setopt extendedglob` somewhere in your config, that's going to happen.

You can also prefix the command with `noglob` or `alias rosrun=noglob rosrun` to get around that.

nkoenig:
I'm also open to using other characters. The `<, >` symbols can't be used either.

mjcarroll:
I personally think that the inconvenience around `[]` with zsh is pretty minor.

 Most of the time, I would imagine that this is going through a launch file of some sort, where it is a non-issue.

PR #45, opened 2019-09-16
[ros2] Port ros_ign_bridge to ROS2

shiveshkhaitan:
Port `ros_ign_bridge` to ROS2. This is a transfer from [ros2_ign](https://github.com/osrf/ros2_ign)

### DIFF
--- a/ros_ign_bridge/CMakeLists.txt
+++ b/ros_ign_bridge/CMakeLists.txt
@@ -11,7 +11,11 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
 endif()
 
 find_package(ament_cmake REQUIRED)
+find_package(geometry_msgs REQUIRED)
+find_package(nav_msgs REQUIRED)
 find_package(rclcpp REQUIRED)
+find_package(rosgraph_msgs REQUIRED)
+find_package(sensor_msgs REQUIRED)
 find_package(std_msgs REQUIRED)
 
 find_package(ignition-msgs4 QUIET REQUIRED)
@@ -19,6 +23,40 @@ set(IGN_MSGS_VER ${ignition-msgs4_VERSION_MAJOR})
 
 find_package(ignition-transport7 QUIET REQUIRED)
 set(IGN_TRANSPORT_VER ${ignition-transport7_VERSION_MAJOR})
+
+include_directories(include)
+
+set(common_sources
+  src/convert_builtin_interfaces.cpp
+  src/builtin_interfaces_factories.cpp
+)
+
+set(bridge_executables
+  parameter_bridge
+  static_bridge
+)
+
+foreach(bridge ${bridge_executables})
+  add_executable(${bridge}
+    src/${bridge}.cpp
+    ${common_sources}
+  )
+  target_link_libraries(${bridge}
+    ignition-msgs${IGN_MSGS_VER}::core
+    ignition-transport${IGN_TRANSPORT_VER}::core
+  )
+  ament_target_dependencies(${bridge}
+    "geometry_msgs"
+    "nav_msgs"
+    "rclcpp"
+    "rosgraph_msgs"
+    "sensor_msgs"
+    "std_msgs"
+  )
+  install(TARGETS ${bridge}
+    DESTINATION lib/${PROJECT_NAME}
+  )
+endforeach()
 
 if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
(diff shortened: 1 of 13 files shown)


ros_ign_bridge/include/ros_ign_bridge/bridge.hpp
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 Open Source Robotics Foundation, Inc.
+// Copyright 2019 Open Source Robotics Foundation, Inc.
```
chapulina:
```suggestion
// Copyright 2018 Open Source Robotics Foundation, Inc.
```

We can leave the original copyright year, since we're porting existing code to ROS 2. This applies to other files too.

ros_ign_bridge/include/ros_ign_bridge/bridge.hpp
```diff
@@ -15,110 +15,96 @@
...
+#include <ros_ign_bridge/builtin_interfaces_factories.hpp>
+
+#include <memory>
+#include <string>
 
 namespace ros_ign_bridge
 {
 
-struct BridgeRosToIgnHandles
+struct BridgeRostoIgnHandles
```
chapulina: Can we leave this with the same name as before? (i.e. `to` -> `To`)

The same goes for other similar variables, like `bridgeRosToIgn`

ros_ign_bridge/include/ros_ign_bridge/bridge.hpp
```diff
@@ -15,110 +15,96 @@
...
-  ros::NodeHandle ros_node,
+  rclcpp::Node::SharedPtr ros_node,
   std::shared_ptr<ignition::transport::Node> ign_node,
   const std::string & ros_type_name,
   const std::string & ros_topic_name,
-  size_t subscriber_queue_size,
   const std::string & ign_type_name,
   const std::string & ign_topic_name,
-  size_t publisher_queue_size)
+  size_t queue_size)
```
chapulina: Let's keep separate queue sizes to reduce the diff from the ROS 1 implementation. I believe that even though it's not used by Ignition yet, it could be used in the future.

ros_ign_bridge/include/ros_ign_bridge/builtin_interfaces_factories.hpp
```diff
@@ -15,39 +15,35 @@
...
-// include ROS messages
-#include <geometry_msgs/Quaternion.h>
-#include <geometry_msgs/Point.h>
-#include <geometry_msgs/Pose.h>
-#include <geometry_msgs/PoseStamped.h>
-#include <geometry_msgs/Transform.h>
-#include <geometry_msgs/TransformStamped.h>
-#include <geometry_msgs/Twist.h>
-#include <geometry_msgs/Vector3.h>
-#include <mav_msgs/Actuators.h>
```
chapulina, line 27: We could leave the `mav_msgs` commented so we remember it still needs to be ported. Let's do that for all the lines where it shows up.

ros_ign_bridge/include/ros_ign_bridge/factory.hpp
```diff
@@ -41,110 +39,81 @@ class Factory : public FactoryInterface
...
   }
 
 protected:
-
   static
   void ros_callback(
-    const ros::MessageEvent<ROS_T const> & ros_msg_event,
-    ignition::transport::Node::Publisher & ign_pub,
-    const std::string &ros_type_name,
-    const std::string &ign_type_name)
```
chapulina, line 113: Can we keep the type names as parameters and use them to print error / info?

ros_ign_bridge/include/ros_ign_bridge/factory.hpp
```diff
@@ -41,110 +39,81 @@ class Factory : public FactoryInterface
...
   void ign_callback(
     const IGN_T & ign_msg,
-    ros::Publisher ros_pub)
+    rclcpp::PublisherBase::SharedPtr ros_pub)
   {
     ROS_T ros_msg;
     convert_ign_to_ros(ign_msg, ros_msg);
-    ros_pub.publish(ros_msg);
+    std::shared_ptr<rclcpp::Publisher<ROS_T>> pub =
+      std::dynamic_pointer_cast<rclcpp::Publisher<ROS_T>>(ros_pub);
```
chapulina, line 126: Can we check if the cast worked before using `pub` below?

ros_ign_bridge/CMakeLists.txt
```diff
@@ -20,6 +24,40 @@ set(IGN_MSGS_VER ${ignition-msgs4_VERSION_MAJOR})
...
+    "std_msgs"
+  )
+  install(TARGETS ${bridge}
+    DESTINATION lib/${PROJECT_NAME}
+  )
+endforeach()
+
 if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
   # TODO(anyone) Enable as files are ported to ROS 2
```
chapulina: You can re-enable these linters now that most files have been ported. You'll need to add an `AMENT_IGNORE` file to the `test` folder so those files are not checked yet.

ros_ign_bridge/include/ros_ign_bridge/factory.hpp
```diff
@@ -15,18 +15,16 @@
...
 
-// include ROS message event
-#include <ros/console.h>
-#include <ros/message.h>
-#include <ros/ros.h>
+// include ROS 2
+#include <rclcpp/rclcpp.hpp>
 
-#include "ros_ign_bridge/factory_interface.hpp"
+#include <ros_ign_bridge/factory_interface.hpp>
```
chapulina: Can we keep the `""` since this is in the same project?

ros_ign_bridge/src/parameter_bridge.cpp
```diff
@@ -12,180 +12,92 @@
...
-// Direction of bridge.
-enum Direction
-{
-  // Both directions.
-  BIDIRECTIONAL = 0,
-  // Only from IGN to ROS
-  FROM_IGN_TO_ROS = 1,
-  // Only from ROS to IGN
-  FROM_ROS_TO_IGN = 2,
-};
```
chapulina: Can we keep all the code for the bridge direction?

chapulina (changes requested):
Looking nice! I tested the static bridge according to [these instructions](https://github.com/osrf/ros2_ign/blob/master/ros2_ign_bridge/README.md#example-3-static-bridge) and it worked fine. I also tested strings and images with the parameter bridge and they worked well :smile: 

Could you update the README with the supported messages and examples?

chapulina (approved):
Works great! I updated the README on cadfc3e8258f10e1b8d02108e0cd770b62438c85.

PR #41, opened 2019-08-20
ROS version agnostic

chapulina:
This is the first step towards making the packages in this repository agnostic to the ROS version, so that we can more cleanly maintain both ROS 1 and ROS 2 integrations.

We're going to retire the recently created [ros2_ign](https://github.com/osrf/ros2_ign) repository and move that code to this repository. It hasn't been decided yet whether the code for ROS 1 and 2 will live in separate branches, or whether we want to split functionality into separate packages and put code specific to each ROS version into one of these smaller packages.

## Commits

* 2ebaa77 : Move files from `ros1_` directories to `ros_` without any changes
* d22e883 : Rename packages and fix tests
* a6f67bf : Rename functions, variables, namespaces, comments, etc.

## Migration plan

- [x] Add Debian renaming rules to ease transition of debian packages (CC @j-rivero )
- [x] Make pull request for SubT to use the new packages
- [x] Rename this repository

PR #46, opened 2019-10-21
[ros_ign -> melodic] Add replaces for each package

chapulina:


chapulina:
This is just syncing with #41. Merging.

PR #47, opened 2019-10-21
[ros_ign -> dashing] Add replaces for each package

chapulina:


chapulina:
Since `ros1_ign` has never been released for ROS 2, this commit can be left out.

PR #48, opened 2019-10-22
[Citadel] Citadel support

chapulina:
* Ignore `ros_ign_point_cloud` until #40 is resolved
* Accept both ign-gazebo 2 and ign-gazebo 3 for the demos
* Other packages not affected, since we're still using `msgs4` and `transport7`

PR #51, opened 2019-11-21
Executable to spawn entities (SDF/URDF)

chapulina:
Builds on top of #48 .

Add a ROS-enabled executable that spawns entities into simulation.

I just added it to the demo packages for now, but it would be nice to eventually move this and other things like `ign_gazebo.launch` to a package which we recommend people to depend on.

Instructions on README.


chapulina:
Moved to #60

PR #53, opened 2019-11-22
Add support for std_msgs/Empty

azeey:


mjcarroll (approved):
LGTM, but probably not necessary to update the tests here, since they are still ROS1.

chapulina (approved):

PR #57, opened 2019-12-09
Update dependencies to Citadel

j-rivero:


ros_ign_image/package.xml
```diff
@@ -7,8 +7,8 @@
 
   <buildtool_depend>catkin</buildtool_depend>
 
-  <depend>ignition-msgs4</depend>
-  <depend>ignition-transport7</depend>
```
chapulina: `ros_ign_bridge` also needs a `package.xml` update

ros_ign_image/CMakeLists.txt
```diff
@@ -16,11 +16,22 @@ find_package(catkin REQUIRED COMPONENTS
   roscpp
```
chapulina: Note that `ros_ign_image` is currently [being ignored](https://github.com/osrf/ros_ign/blob/dashing/ros_ign_image/COLCON_IGNORE) for ROS 2. But I think it doesn't hurt to update the cmake to match melodic's.

ros_ign_image/package.xml
j-rivero: I think that we have now them updated after the merge of https://github.com/osrf/ros_ign/pull/58

ros_ign_image/CMakeLists.txt
```diff
@@ -48,6 +48,9 @@ target_link_libraries(${executable}
   ignition-msgs${IGN_MSGS_VER}::core
   ignition-transport${IGN_TRANSPORT_VER}::core
 )
+install(TARGETS ${executable}
+        DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
+)
```
chapulina: This is already below on line 62, and has been updated for the `ament` pattern

ros_ign_bridge/package.xml
```diff
@@ -12,6 +12,8 @@
 
   <buildtool_depend>ament_cmake</buildtool_depend>
 
+  <depend>ignition-msgs5</depend>
+  <depend>ignition-transport8</depend>
```
chapulina, line 16: Nice catch :+1:

chapulina (changes requested):

chapulina (approved):

PR #58, opened 2019-12-11
[WIP] Port ign_ros_gazebo_demos to ROS2

j-rivero:
First effort to port ros_gazebo_demos to ROS2

All launch XML files have been converted to python. All components of the old XML launch configuration are present in the current python files.  

There still major problems:
 * The bridge fails to me on every launch file I ported, not sure why probably something stupid, with these kind of errors:
```
[parameter_bridge-2] Failed to create a bridge for topic [/rgbd_camera/image] with ROS2 type [sensor_msgs/Image] and Ignition Transport type [ignition.msgs.Image]
```
 * Some files can not be ported yet until other packages in the repo are ported (comment done in launch file).
 * ign gazebo parameters -r -v present in some files were not ported. I was unable to pass these paremeters correctly.
 * Seems like all rviz configurations needs porting and some features are missing. None of them worked out of the box.
 * We need to investigate the problems with ignition-gazebo3 cmake (complaining about DART cmake bad used). Code is commented out by now.


chapulina:
I added a few changes in 6b5063ec53406a87cb44fde0ef81e48601e8acb5, addressing demos down to differential drive.

> The bridge fails to me on every launch file I ported

I think the main problem is the missing `msg` namespace, i.e. `geometry_msgs/Twist` in ROS 1 becomes `geometry_msgs/msg/Twist` in ROS 2. Besides that, it looks like we still need to port other messages, like `FluidPressure`.

> Seems like all rviz configurations needs porting and some features are missing. None of them worked out of the box.

I think we can just regenerate them by manually adding the displays and saving the new config. I'm not sure if the configs are meant to be portable from ROS 1 to ROS 2, so I think it's ok to do this for all demos.

chapulina (approved):
Thanks @j-rivero and @JShep1 for the collective effort! All packages and demos that could be migrated have been migrated!

Some are blocked by `ros_ign_point_cloud` (also a problem on ROS 1), and the `air_pressure` demo is blocked because we haven't migrated the `FluidPressure` message yet.

PR #59, opened 2019-12-11
Port ign image to ros2

JShep1:


chapulina:
Thanks, @JShep1 , works great! I just had to add `ament_export_include_directories(include)` for it to compile on my workspace.

I also pushed a few smaller changes on 56a2ac9a556f8b5aa778b4b55368ed242245e444:

* Added `AMENT_IGNORE` to the `test` directory so the linters don't look at it just yet
* Fixed all linter issues. Those can be checked running `colcon test --packages-select-regex ros_ign_image` and then looking at `log/latest_test/ros_ign_image/streams.log`, for example.
* Kept the test cmake code commented out so we don't lose track of it.

I'll merge this into #58 and then we can get it all into `dashing` at once :smile:

chapulina (approved):

PR #64, opened 2020-03-24
Add Float64 to ignition Double conversions.  This is needed for suppo…

realdealneil:
Add Float64 to ignition Double conversions.  This is needed for supporting joint controller (such as for a pan/tilt gimbal).  Forthcoming dependent pull requests will be issued on OSRF SubT project to support a pan/tilt gimbal.


chapulina (approved):
LGTM! Please consider also porting this PR to ROS 2 :smile:

PR #65, opened 2020-03-27
Ignore ros-args in parameter bridge

luca-della-vedova:
Starting from eloquent, the --ros-args flag becomes mandatory and, according to the [design document](http://design.ros2.org/articles/ros_command_line_arguments.html), even an empty `--ros-flags` is valid.

When embedding the parameter_bridge in a launch file in eloquent, i.e. in the following way (forwarding ignition clock to ROS2):

``` 
    <node pkg="ros_ign_bridge" exec="parameter_bridge"
        args="/clock@rosgraph_msgs/msg/Clock[ignition.msgs.Clock"
    />
```

ros_ign_bridge crashes because it doesn't expect the --ros-args flag:

`[ERROR] [parameter_bridge-1]: process has died [pid 1398, exit code 255, cmd '/home/luca/demos_ws/install/ros_ign_bridge/lib/ros_ign_bridge/parameter_bridge /clock@rosgraph_msgs/msg/Clock[ignition.msgs.Clock --ros-args'].
` 

This PR ignores every command line argument after the --ros-args (since it's not used anyway in the bridge), fixing the crash.

### DIFF
--- a/ros_ign_bridge/CMakeLists.txt
+++ b/ros_ign_bridge/CMakeLists.txt
@@ -20,7 +20,7 @@ find_package(std_msgs REQUIRED)
 
 # Citadel
 find_package(ignition-transport8 QUIET)
-if (ignition-transport8_FOUND)
+if(ignition-transport8_FOUND)
   set(IGN_TRANSPORT_VER ${ignition-transport8_VERSION_MAJOR})
 
   find_package(ignition-msgs5 REQUIRED)
--- a/ros_ign_bridge/src/parameter_bridge.cpp
+++ b/ros_ign_bridge/src/parameter_bridge.cpp
@@ -24,6 +24,7 @@
 #include <list>
 #include <memory>
 #include <string>
+#include <vector>
 
 // Direction of bridge.
 enum Direction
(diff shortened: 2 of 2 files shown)


chapulina:
Technically, user-defined arguments could come after a `--` after `--ros-args`. Should we handle that case? i.e. ignore all args between `--ros-args` and `--`?

luca-della-vedova:
True, the specific use case was for use in launch files but it makes sense to fix all the cases.
Since it was becoming quite ugly if we kept the same structure (i.e. we would have needed to "change" the iterator in two places or edit the vector while it is being iterated on) I created a new function that to preprocess the arguments and convert them to a vector of strings for easier iteration.

Tested with the following commands:
Original use case:
`ros2 run ros_ign_bridge parameter_bridge /clock@rosgraph_msgs/msg/Clock@ignition.msgs.Clock --ros-args`

And recommended use case:
`ros2 run ros_ign_bridge parameter_bridge --ros-args test1 -test2 __test3 -- clock@rosgraph_msgs/msg/Clock@ignition.msgs.Clock`

chapulina (approved):
LGTM! @luca-della-vedova , could you just fix the `lint_cmake` and `uncrustify` test failures before merging? Thanks!

We gotta check why Travis didn't mark the build as failed - CC :man_farmer:  @j-rivero

PR #66, opened 2020-04-06
Corrected error in launch with wrong package name in the create demo for ROS1 Melodic

RDaneelOlivav:
Hi,

I came across this minor issue while executing the demos.
Now the three models appear without issue.


chapulina (approved):
Well spotted, thanks for the fix!

PR #67, opened 2020-04-17
Add ignition::msgs::Pose_V to tf2_msgs::TFMessage conversion

iche033:
For sending a group of transform stamped messages 

Also fixed test that was failing at the `double` msg type.

### DIFF
--- a/ros_ign_bridge/src/convert_builtin_interfaces.cpp
+++ b/ros_ign_bridge/src/convert_builtin_interfaces.cpp
@@ -376,6 +376,41 @@ convert_ign_to_ros(
 template<>
 void
 convert_ros_to_ign(
+  const tf2_msgs::TFMessage & ros_msg,
+  ignition::msgs::Pose_V & ign_msg)
+{
+  ign_msg.clear_pose();
+  for (auto const &t : ros_msg.transforms)
+  {
+    auto p = ign_msg.add_pose();
+    convert_ros_to_ign(t, *p);
+  }
+
+  if (!ros_msg.transforms.empty())
+  {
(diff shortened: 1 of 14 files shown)


ros_ign_bridge/src/convert_builtin_interfaces.cpp
```diff
@@ -373,6 +373,41 @@ convert_ign_to_ros(
...
+  const ignition::msgs::Pose_V & ign_msg,
+  tf2_msgs::TFMessage & ros_msg)
+{
+  ros_msg.transforms.clear();
+  for (auto const &p : ign_msg.pose())
+  {
+    geometry_msgs::TransformStamped tf;
+    convert_ign_to_ros(p, tf);
+    ros_msg.transforms.push_back(tf);
+  }
```
nkoenig, line 408:
```suggestion
  }
  if (!ros_msg.transforms.empty())
  {
    convert_ign_to_ros(ign_msg.header(),
        ros_msg.transforms[0].header);
  }
```

nkoenig (commented):
Just one suggestion.

nkoenig (approved):

ros_ign_bridge/src/convert_builtin_interfaces.cpp
iche033: just saw this comment. For `convert_ign_to_ros`, the transforms should have timestamps set already so we shouldn't have to set it. I had to do this in `convert_ros_to_ign` because ignition::msgs::Pose_V has a header field but `tf2_msgs/TFMessage` does not so I set ign msgs's header to the header of the first transform mainly to get the timestamp.

PR #68, opened 2020-04-18
Travis CI for multiple versions with travis

chapulina:
This makes Travis trigger 2 jobs in parallel, one for Blueprint and another for Citadel, so we're sure that this branch works with both.


j-rivero:
great!

j-rivero (approved):

PR #71, opened 2020-05-13
Fix gazebo version in package.xml files

nkoenig:
Signed-off-by: Nate Koenig <nate@openrobotics.org>

PR #72, opened 2020-05-13
Choose collection based on environment variable

chapulina:
Since #48, we support both Blueprint and Citadel at the CMake level, but `package.xml` has only been supporting Citadel.

Then on #71, a couple of package-level dependencies were reverted to Blueprint to get a release out.

This pull request leverages the `condition` (see [REP 149](https://www.ros.org/reps/rep-0149.html)) argument to support both collections based on the `IGNITION_VERSION` environment variable, defaulting to Citadel when the variable is not present.


ros_ign_gazebo/package.xml
```diff
@@ -8,6 +8,10 @@
   <buildtool_depend>catkin</buildtool_depend>
 
   <depend>libgflags-dev</depend>
-  <exec_depend>ignition-gazebo2</exec_depend>
+
+  <!-- Default to citadel, support blueprint -->
+  <exec_depend condition="$IGNITION_VERSION == blueprint">ignition-gazebo2</depend>
+  <exec_depend condition="$IGNITION_VERSION != blueprint">ignition-gazebo3</depend>
```
iche033: `exec_depend` -> `depend`

iche033 (changes requested):

ros_ign_gazebo/package.xml
chapulina: Done 
197af4a

iche033 (approved):

chapulina:
@j-rivero , I think this will break blooming Blueprint-compatible packages unless we expose the Ignition version to `release-tools/bloom/release-bloom.py`. It shouldn't be as simple as setting the `$IGNITION_VERSION` variable before calling that script, I think. The dependencies are fetched inside the Jenkins node, right?

j-rivero:
>  It shouldn't be as simple as setting the `$IGNITION_VERSION` variable before calling that script, I think.

If I'm not wrong, the rosdep keys are resolved at the time of running bloom which usually happens in the developer local system. We need to integrate the `$IGNITION_VERSION` in our releasing tools but probably it does not affect to Jenkins scripts since at that time the only thing used are debian package names generated by bloom.

j-rivero:
> defaulting to Citadel when the variable is not present.

If `IGNITION_VERSION` is not defined, my system is trying to get Blueprint dependencies (see colcon log about ignition-gazebo2) instead of defaulting to Citadel. Am I doing something wrong?

```
jrivero@nium ros_ign_ws $ unset IGNITION_VERSION &&  MAKEJOBS="-j1" colcon build --executor sequential --event-handler console_direct+  | grep ignition-gazebo2
CMake Error at CMakeLists.txt:26 (find_package):
  By not providing "Findignition-gazebo2.cmake" in CMAKE_MODULE_PATH this
  project has asked CMake to find a package configuration file provided by
  "ignition-gazebo2", but CMake did not find one.
```

chapulina:
> If IGNITION_VERSION is not defined, my system is trying to get Blueprint dependencies 

That error is coming in the CMake level, not the package level. The CMake code is not aware of the new env var and will always try Citadel first. Maybe we should change that :bulb: 

---

The way I tested this PR with colcon was:

1. Drop `ros_ign` in a workspace with other Ignition libs, i.e.

```
src/
|-- ign-cmake
|-- ign-common
|-- ign-fuel-tools
|-- ign-gazebo
|-- ign-gui
|-- ign-launch
|-- ign-math
|-- ign-msgs
|-- ign-physics
|-- ign-plugin
|-- ign-rendering
|-- ign-sensors
|-- ign-tools
|-- ign-transport
|-- ros_ign 
`-- sdformat
```

2. Use `colcon graph` to check the inter-dependencies inside the workspace:

`colcon graph --dot | grep "ros_ign_.*->.*ignition"`

If you have blueprint dependencies in the workspace and `IGNITION_VERSION=blueprint`:

```
  "ros_ign_gazebo_demos" -> "ignition-gazebo2" [color="#0000ff:#ff0000"];
  "ros_ign_gazebo" -> "ignition-gazebo2" [color="#0000ff:#ff0000"];
  "ros_ign_image" -> "ignition-transport7" [color="#0000ff:#ff0000"];
  "ros_ign_image" -> "ignition-msgs4" [color="#0000ff:#ff0000"];
  "ros_ign_bridge" -> "ignition-transport7" [color="#0000ff:#ff0000"];
  "ros_ign_bridge" -> "ignition-msgs4" [color="#0000ff:#ff0000"];
```

If the variable is anything other than blueprint, the command won't show anything.

You can also try the other way around in a citadel workspace.

chapulina:
@j-rivero , I updated the CMake code to also check the environment variables.

I also went ahead and tackled #90 while documenting the `IGNITION_VERSION` variable.

This is ready for review again.

j-rivero:
Code looks good, compilation against Blueprint and Citadel works as expected and a quick smoke test seems to be able to load things in memory. Nice work.

j-rivero (approved):

PR #80, opened 2020-06-08
Added ros_ign_gazebo for ros2

ahcorde:
This package contains things that make it convenient to integrate ROS with Ignition, such as:

 - Launch files
 - ROS-enabled executables

## Run ignition Gazebo

```bash
ros2 launch ros_ign_gazebo ign_gazebo.launch.py ignition_args:="shapes.sdf"

# or

ros2 launch ros_ign_gazebo ign_gazebo_gui.launch.py 
ros2 launch ros_ign_gazebo ign_gazebo_server.launch.py ignition_args:="shapes.sdf"
```

```bash
ros2 run ros_ign_gazebo ign_gazebo shapes.sdf
```

### Spawn entities

```
ros2 run ros_ign_gazebo create -world default -file 'https://fuel.ignitionrobotics.org/1.0/openrobotics/models/Gazebo'
```

By the way this command is adding the entity to ignition gazebo, but the mesh doesn't appear.

-------

**Note**: I created the branch `ros2` to keep it as the devel branch. 

Signed-off-by: ahcorde <ahcorde@gmail.com>


ros_ign_gazebo/launch/ign_gazebo_gui.launch.py
```diff
@@ -0,0 +1,35 @@
...
+from launch import LaunchDescription
+from launch.actions import ExecuteProcess
+from launch.actions import Shutdown
+
+
+def generate_launch_description():
+
+    ros_ign_gazebo = get_package_prefix('ros_ign_gazebo')
+
+    cmd = [[ros_ign_gazebo + '/lib/ros_ign_gazebo/ign_gazebo', ' ', '-g', ' ']]
```
chapulina: Can we also propagate `ignition_args` to the GUI? It also accepts some options like `-v` and `--gui-config`

ros_ign_gazebo/scripts/ign_gazebo
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+# TODO This approach has way too many intermediate steps:
```
chapulina: I think for ROS 2 it's possible to do something _a bit_ more direct, like this:

https://github.com/ignitionrobotics/ros_ign/blob/82354e4923c5bdfbe239b60654fc765567b74a9a/ros_ign_gazebo_demos/launch/ign_gazebo.launch.py#L27-L37

ros_ign_gazebo/scripts/ign_gazebo
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
chapulina: For some reason this file was giving me `Permission denied`. I had to do `chmod +x`

ros_ign_gazebo/launch/ign_gazebo.launch.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def generate_launch_description():
+
+    return LaunchDescription([
+        DeclareLaunchArgument('gui', default_value='true',
+                              description='Set to "false" to run headless.'),
+
+        DeclareLaunchArgument('server', default_value='true',
+                              description='Set to "false" not to run gzserver.'),
```
chapulina: I'm on the fence about splitting this into `gui` and `server`. Ignition Gazebo already supports this internally with the `-s` and `-g` flags. I feel like we could be duplicating the efforts here. I don't feel too strongly about it, but I have a feeling that `ign-gazebo` should know better how to manage the lifecycle for the GUI and server processes.

chapulina (changes requested):
Thanks for tackling this so quickly, @ahcorde ! I tried it with Eloquent and it worked for me. It would be nice to document the supported versions on the README. Also, I'm not sure why Travis wasn't triggered :monocle_face: I see that it's supposed to install `dashing`.

> I created the branch ros2 to keep it as the devel branch.

Would that be for `eloquent` and `foxy`?. Could we target this PR at `dashing`? Usually on Ignition we target PRs at the lowest compatible version and merge changes forward (as opposed to how on ROS commits are cherry-picked backwards).

---

Some other requests that could come in follow-up PRs:

* Could you add the package to `ros_ign/package.xml`?
* Can we use the new launch file on `ros_ign_gazebo_demos`? We should still leave the existing one there for a while in case someone is using it, but we shouldn't encourage its use anymore.

ros_ign_gazebo/launch/ign_gazebo_gui.launch.py
ahcorde: Added 
 - `ignition_server_args`
 - `ignition_gui_args`

ros_ign_gazebo/scripts/ign_gazebo
chapulina: Did you give any thought to this? Is there a downside to calling `ign gazebo` directly from the launch file? I think this bash script adds an unnecessary intermediate step.

ros_ign_gazebo/launch/ign_gazebo.launch.py
chapulina: Another argument for a single entry point is that now we have arguments split between gui and server args. Users of `ign gazebo` don't need to think about these differences, so I think users of these launch files also shouldn't have to. Another downside is that arguments that apply to both, such as `-v 4`, would need to be duplicate.

The more I think the more I'm convinced that we should let `ign gazebo` handle the GUI and server processes.

What do you think?

ros_ign_gazebo/package.xml
```diff
@@ -0,0 +1,26 @@
...
+
+  <license>Apache 2.0</license>
+
+  <author>Alejandro Hernandez</author>
+
+  <buildtool_depend>ament_cmake</buildtool_depend>
+
+  <depend>rclcpp</depend>
+  <depend>ignition-math6</depend>
+  <depend>ignition-msgs</depend>
```
chapulina: We won't have generic rosdep keys for Ignition libraries, so we always need to specify the version:

```suggestion
  <depend>ignition-msgs5</depend>
```

ros_ign_gazebo/package.xml
```diff
@@ -0,0 +1,26 @@
...
+  <depend>ignition-transport</depend>
```
chapulina:
```suggestion
  <depend>ignition-transport8</depend>
```

chapulina (commented):
Everything is working for me, I'd just like to hear your thoughts about:

* not splitting gui / server launch files
* not using a bash script

ahcorde:
> not splitting gui / server launch files

I don't have a strong opinion here. I created two launch files just to keep the same idea that we have in Gazebo and  provide a launch file to run ign-gazebo headless mode (without reviewing/knowing the arguments). 

But Gazebo is a different case where we have two executables in Ignition we only have one. The idea of duplicate the arguments sounds weird and it's going to create issues in the user side.

I think is a good idea to handle everything in the same launch file.

> not using a bash script

the bash script is not adding any value I can remove it

chapulina:
> I'm not sure why Travis wasn't triggered

This org hadn't granted permissions. There's a build running now: https://travis-ci.org/github/ignitionrobotics/ros_ign/builds/698585638

chapulina (approved):
I skipped one more key on CI in 9d8ab84 and fixed some linters on 26025a3. I think this is good to merge if Travis comes back green.

PR #84, opened 2020-06-15
Avoid the use of --ros-args arguments outside ros

ahcorde:
The Handler class only accepts non-ros arguments. This PR avoid the use `--ros-args` arguments with this class

Signed-off-by: ahcorde <ahcorde@gmail.com>


chapulina (approved):
Thanks!

PR #82, opened 2020-06-09
Updated launch file to use ros_ign_gazebo

ahcorde:
This PR builds on top of this other PR #80

This PR update the lanch file to use the launch file inside ros_ign_gazebo

Signed-off-by: ahcorde <ahcorde@gmail.com>

PR #85, opened 2020-06-24
Use new ros_ign_gazebo package on ROS 2 demos

chapulina:
An alternative to #82 that works with Dashing.


ros_ign_gazebo_demos/launch/depth_camera.launch.py
```diff
@@ -62,7 +63,7 @@ def generate_launch_description():
     # FIXME: need a SDF file (depth_camera.sdf) inside ros_ign_point_cloud
     return LaunchDescription([
         DeclareLaunchArgument(
-          'args',
+          'ign_args',
           #default_value=['depth_camera.sdf'],
```
ahcorde, line 68: you need to define at least a default value
```suggestion
          default_value=[],
          #default_value=['depth_camera.sdf'],
```

ahcorde (changes requested):

ros_ign_gazebo_demos/launch/depth_camera.launch.py
chapulina: Ah thanks! This file is not working right now but it's good to take this out of the way.

ahcorde (approved):

PR #94, opened 2020-07-21
Update Foxy CI

chapulina:
Install Citadel dependencies from rosdep keys.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: required
 
 env:
   matrix:
-    - DOCKER_IMAGE=ubuntu:20.04 IGNITION_VERSION="citadel" ROS_DISTRO="foxy"
+    - DOCKER_IMAGE=ubuntu:20.04 ROS_DISTRO="foxy"
 
 services:
   - docker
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -2,7 +2,7 @@ sudo: required
 
 env:
   matrix:
-    - DOCKER_IMAGE=ubuntu:20.04 IGN_COLLECTION="citadel" ROS_DISTRO="foxy"
+    - DOCKER_IMAGE=ubuntu:20.04 IGNITION_VERSION="citadel" ROS_DISTRO="foxy"
```
j-rivero: if we are going to support only one version, I suggest to remove it or add a new entry to the matrix without `IGNITION_VERSION`
```suggestion
    - DOCKER_IMAGE=ubuntu:20.04 ROS_DISTRO="foxy"
```

j-rivero (approved):
nice to remove all those hardcoded packages! Minor suggestion for CI, looks good.

.travis.yml
chapulina: Removed: 5bf9b5c8e3350955191380d56c1f30ca4f8d94db

PR #95, opened 2020-07-22
Install only what's necessary, rename builtin_interfaces

chapulina:
Moved some headers from `include` to `src` because they don't need to be installed.

I also removed `builtin_interfaces` from the names of a few files, because I don't think they only refer to [this package](https://github.com/ros2/rcl_interfaces/tree/master/builtin_interfaces) anymore.

:warning: Due to all the moved files, please don't squash-merge.


ros_ign_bridge/src/convert.cpp
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <ros_ign_bridge/convert_builtin_interfaces.hpp>
+#include <ros_ign_bridge/convert.hpp>
```
j-rivero: NitPick: 
```suggestion
#include "ros_ign_bridge/convert.hpp"
```

j-rivero (approved):
minor nitpick about code style (not sure if its correct).

chapulina:
Re-triggering Travis (for some reason I don't have permissions to do it through their UI)

chapulina:
> Due to all the moved files, please don't squash-merge.

... and I squashed 🤦

chapulina:
Phew, ok, we're good, GitHub is smart and the moved files kept their history: https://github.com/ignitionrobotics/ros_ign/blame/dashing/ros_ign_bridge/src/factories.hpp

PR #96, opened 2020-07-22
Install only what's necessary, rename builtin_interfaces

chapulina:
ROS 1 version of #95 

Moved some headers from `include` to `src` because they don't need to be installed.

I also removed `builtin_interfaces` from the names of a few files, because I don't think they only refer to [this package](https://github.com/ros2/rcl_interfaces/tree/master/builtin_interfaces) anymore.

:warning: Due to all the moved files, please don't squash-merge.


ros_ign_bridge/include/ros_ign_bridge/convert.hpp
```diff
@@ -44,10 +44,10 @@
 #include <std_msgs/String.h>
 #include <tf2_msgs/TFMessage.h>
 
-// include Ignition builtin messages
+// include Ignition messages
 #include <ignition/msgs.hh>
 
-#include "ros_ign_bridge/convert_decl.hpp"
+#include <ros_ign_bridge/convert_decl.hpp>
```
j-rivero: Not sure about the style. Are we using `<...>` for headers in the same package?

j-rivero (approved):
minor nitpick, otherwise compiles and looks good.

ros_ign_bridge/include/ros_ign_bridge/convert.hpp
chapulina: good point: acbe7d78fa23145ede636b58edbf89d7c5def704

PR #102, opened 2020-08-19
Add pkg-config as a buildtool dependency

chapulina:
@nuclearsandwich kindly helped me investigate why the `ros_ign` packages are failing to build for Foxy and this is a promising band-aid solution.

### DIFF
--- a/ros_ign_bridge/package.xml
+++ b/ros_ign_bridge/package.xml
@@ -11,6 +11,7 @@
   <author>Shivesh Khaitan</author>
 
   <buildtool_depend>ament_cmake</buildtool_depend>
+  <buildtool_depend>pkg-config</buildtool_depend>
 
   <depend>ignition-msgs5</depend>
   <depend>ignition-transport8</depend>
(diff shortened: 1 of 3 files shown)


nuclearsandwich:
I think you also need this in ros_ign_gazebo based on the failures here: http://build.ros2.org/view/Fbin_uF64/job/Fbin_uF64__ros_ign_gazebo__ubuntu_focal_amd64__binary/

nuclearsandwich:
If ign-cmake was a ROS package, the common practice would be to add a buildtool_export_depend on pkg-config so that any package downstream of ign-cmake got the dependency needed to use its bundled cmake modules.

It's not a ROS package so while that possibility is entertained this PR moves forward with a local solution to include pkg-config as a buildtool dependency in the packages which use ign-cmake bundled modules that expect pkg-config to be available.

chapulina:
> I think you also need this in ros_ign_gazebo

You're right, I forgot that also links against `ign-transport`. 02be7a7

nuclearsandwich (approved):
This should at least get us building as a temporary workaround if not the desired approach.

PR #103, opened 2020-08-20
Add support for Dome

luca-della-vedova:
Just adding an additional check in the CMakeLists to look for the Dome libraries, the order of priority is:

Citadel
Dome
Blueprint (compile error if none found)

I didn't update ros_ign_point_cloud since it never supported Citadel in the first place, not sure this is intended?

### DIFF
--- a/ros_ign_bridge/CMakeLists.txt
+++ b/ros_ign_bridge/CMakeLists.txt
@@ -19,7 +19,6 @@ find_package(sensor_msgs REQUIRED)
 find_package(std_msgs REQUIRED)
 
 find_package(ignition-transport8 QUIET)
-find_package(ignition-transport9 QUIET)
 # Citadel
 if(ignition-transport8_FOUND)
   set(IGN_TRANSPORT_VER ${ignition-transport8_VERSION_MAJOR})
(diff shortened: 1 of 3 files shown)


ahcorde (approved):
LGTM

ros_ign_bridge/CMakeLists.txt
```diff
@@ -18,14 +18,22 @@ find_package(rosgraph_msgs REQUIRED)
 find_package(sensor_msgs REQUIRED)
 find_package(std_msgs REQUIRED)
 
-# Citadel
 find_package(ignition-transport8 QUIET)
+find_package(ignition-transport9 QUIET)
```
chapulina: In order to avoid problems in case both `transport8` and `transport9` are available, can we move the find call for `transport9` into a block where `transport8` is not found?

chapulina (changes requested):

ros_ign_bridge/CMakeLists.txt
luca-della-vedova: [Done](https://github.com/ignitionrobotics/ros_ign/pull/103/commits/b40e4ff37b2364c91d5a18bbd6c61a734da750af)

chapulina (approved):

chapulina:
Ouch, after I merged I remembered that the `package.xml`s could also be updated:

https://github.com/ignitionrobotics/ros_ign/blob/acff2f21a57a14191d120572157deec97f3a38c2/ros_ign_gazebo/package.xml#L13-L14

That would help a colcon workspace get the correct dependencies.

PR #107, opened 2020-08-29
[dashing] Update releases

chapulina:
* Update list on README to include latest releases
* Update RELEASING info (from `melodic` branch)
* Update CI so it doesn't use `[trusted=yes]`
* Clean up support for both Blueprint and Citadel, using env var

### DIFF
--- a/ros_ign_gazebo_demos/package.xml
+++ b/ros_ign_gazebo_demos/package.xml
@@ -8,9 +8,9 @@
   <buildtool_depend>ament_cmake</buildtool_depend>
 
   <!-- Blueprint -->
-  <depend condition="$IGNITION_VERSION == blueprint">ignition-gazebo2</depend>
+  <exec_depend condition="$IGNITION_VERSION == blueprint">ignition-gazebo2</exec_depend>
   <!-- Citadel (default) -->
-  <depend condition="$IGNITION_VERSION != blueprint">ignition-gazebo3</depend>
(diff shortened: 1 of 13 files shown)


ros_ign_gazebo_demos/package.xml
```diff
@@ -7,7 +7,11 @@
 
   <buildtool_depend>ament_cmake</buildtool_depend>
 
-  <depend>ignition-gazebo3</depend>
+  <!-- Blueprint -->
+  <depend condition="$IGNITION_VERSION == blueprint">ignition-gazebo2</depend>
```
ahcorde: I think this could be a `<exec_depend>`.

ahcorde (approved):
LGTM just a minor comment

ros_ign_gazebo_demos/package.xml
chapulina: Good point 6a24b28

ahcorde:
travis is failing here:

```
Error(s) in package '/code/ros_ign_bridge/package.xml':
The manifest contains invalid XML:
not well-formed (invalid token): line 28, column 44
```

but I have tested it locally and I can't reproduce the error

chapulina:
> The manifest contains invalid XML:

I had committed some stuff that I was playing with by mistake, it should be fixed now.

ahcorde (approved):

PR #120, opened 2020-10-27
Changed for loading xml from ROS param(#119)

### LINKED ISSUE #119: Failed to load xml from ROS param [ros_ign_gazebo create]
## Test environment
- Ubuntu 20.04
- ROS2 Foxy
- ignition citadel

## Test launch file
```
<?xml version="1.0"?>
<launch>
  <group>
    <include file="$(find-pkg-share ros_ign_gazebo)/launch/ign_gazebo.launch.py">
      <arg name="ign_args" value="empty.sdf" />
    </include>
  </group>

  <node pkg="ros_ign_gazebo" exec="create" args="-world empty -param robot_description" output="screen">
    <param name="robot_description" value="$(command 'cat $(find-pkg-share test_robot_ign_gazebo)/urdf/my_robot.urdf')" />
  </node>
</launch>
```
This failed for "Failed to get XML from param".

### Before
https://github.com/ignitionrobotics/ros_ign/blob/94a719adcf6005fae657a174489cb99349c33ced/ros_ign_gazebo/src/create.cpp#L44-L52

### After
```
   ros::init(_argc, _argv, "ign_create"); 
   ros::NodeHandle nh; 

   gflags::AllowCommandLineReparsing(); 
   gflags::SetUsageMessage( 
       R"(Usage: create -world [arg] [-file FILE] [-param PARAM] [-string STRING] 
                        [-name NAME] [-X X] [-Y Y] [-Z Z] [-Roll ROLL] 
                        [-Pitch PITCH] [-Yaw YAW])"); 
   gflags::ParseCommandLineFlags(&_argc, &_argv, true);   
```
This change works good for me.

chama1176:
I have fixed #119

### DIFF
--- a/ros_ign_gazebo/src/create.cpp
+++ b/ros_ign_gazebo/src/create.cpp
@@ -40,7 +40,6 @@ DEFINE_double(Y, 0, "Yaw component of initial orientation, in radians.");
 // If these are not needed, just use the `ign service` command line instead.
 int main(int _argc, char ** _argv)
 {
-
   rclcpp::init(_argc, _argv);
   auto ros2_node = rclcpp::Node::make_shared("ros_ign_gazebo");
 
--- a/ros_ign_gazebo/src/create.cpp
+++ b/ros_ign_gazebo/src/create.cpp
@@ -40,15 +40,16 @@ DEFINE_double(Y, 0, "Yaw component of initial orientation, in radians.");
 // If these are not needed, just use the `ign service` command line instead.
 int main(int _argc, char ** _argv)
 {
+
+  rclcpp::init(_argc, _argv);
+  auto ros2_node = rclcpp::Node::make_shared("ros_ign_gazebo");
+
   gflags::AllowCommandLineReparsing();
   gflags::SetUsageMessage(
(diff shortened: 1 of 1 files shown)


ros_ign_gazebo/src/create.cpp
```diff
@@ -40,16 +40,17 @@ DEFINE_double(Y, 0, "Yaw component of initial orientation, in radians.");
 // If these are not needed, just use the `ign service` command line instead.
 int main(int _argc, char ** _argv)
 {
+
```
chapulina: Can you fix this warning?

```
/code/ros_ign_gazebo/src/create.cpp:43:  Redundant blank line at the start of a code block should be deleted.  [whitespace/blank_line] [2]
```

chapulina (changes requested):
Thank you for the PR, it looks good to me once the warning is fixed.

ros_ign_gazebo/src/create.cpp
chama1176: Thank you for checking.
I' ll fix it.

chapulina (approved):
I haven't tried the exact setup on #119, but I checked that the basic behaviour still works the same way, so I don't see a problem getting this in.

Thank you for the PR!

PR #125, opened 2020-12-08
[ros2] Add default value for plugin path in ign_gazebo launch script 

AndrejOrsula:
In https://github.com/ignitionrobotics/ros_ign/pull/122, I have unfortunately introduced a regression in case `IGN_GAZEBO_SYSTEM_PLUGIN_PATH` was not previously set. My mistake... I forgot to test such use case prior to submitting that PR. 

This PR should fix it by having a default value in case it is not set. The same logic is applied also for `LD_LIBRARY_PATH` because a similar issue could occur for non-Linux systems. I am not that familiar with Python, so I am unsure if there is a more standard way of doing this.

### DIFF
--- a/ros_ign_gazebo/launch/ign_gazebo.launch.py
+++ b/ros_ign_gazebo/launch/ign_gazebo.launch.py
@@ -23,8 +23,9 @@ from launch.substitutions import LaunchConfiguration
 
 
 def generate_launch_description():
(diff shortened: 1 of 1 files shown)


chapulina (approved):
Oh thanks for catching that and for the fix! I'm also not a Python expert but this looks reasonable to me :+1:

PR #126, opened 2020-12-08
Add dome + melodic to CI

nkoenig:
Signed-off-by: Nate Koenig <nate@openrobotics.org>

### DIFF
--- a/ros_ign_bridge/CMakeLists.txt
+++ b/ros_ign_bridge/CMakeLists.txt
@@ -28,7 +28,7 @@ if ("$ENV{IGNITION_VERSION}" STREQUAL "blueprint")
   set(IGN_MSGS_VER ${ignition-msgs4_VERSION_MAJOR})
 
   message(STATUS "Compiling against Ignition Blueprint")
-else()
+elseif("$ENV{IGNITION_VERSION}" STREQUAL "citadel")
   find_package(ignition-transport8 REQUIRED)
   set(IGN_TRANSPORT_VER ${ignition-transport8_VERSION_MAJOR})
 
@@ -36,6 +36,14 @@ else()
   set(IGN_MSGS_VER ${ignition-msgs5_VERSION_MAJOR})
 
   message(STATUS "Compiling against Ignition Citadel")
+else()
+  find_package(ignition-transport9 REQUIRED)
+  set(IGN_TRANSPORT_VER ${ignition-transport9_VERSION_MAJOR})
+
+  find_package(ignition-msgs6 REQUIRED)
+  set(IGN_MSGS_VER ${ignition-msgs6_VERSION_MAJOR})
+
+  message(STATUS "Compiling against Ignition Dome")
 endif()
 
 catkin_package()
--- a/ros_ign_bridge/CMakeLists.txt
+++ b/ros_ign_bridge/CMakeLists.txt
@@ -135,4 +135,3 @@ foreach(test_subscriber ${test_subscribers})
     ignition-transport${IGN_TRANSPORT_VER}::core
   )
 endforeach(test_subscriber)
-
--- a/ros_ign_bridge/CMakeLists.txt
+++ b/ros_ign_bridge/CMakeLists.txt
@@ -19,7 +19,7 @@ find_package(catkin REQUIRED COMPONENTS
                std_msgs
                tf2_msgs)
 
-# Default to Citadel, support Blueprint
+# Default to Dome, support Citadel and Blueprint
 if ("$ENV{IGNITION_VERSION}" STREQUAL "blueprint")
   find_package(ignition-transport7 REQUIRED)
   set(IGN_TRANSPORT_VER ${ignition-transport7_VERSION_MAJOR})
--- a/ros_ign_bridge/CMakeLists.txt
+++ b/ros_ign_bridge/CMakeLists.txt
@@ -135,3 +135,4 @@ foreach(test_subscriber ${test_subscribers})
     ignition-transport${IGN_TRANSPORT_VER}::core
   )
 endforeach(test_subscriber)
+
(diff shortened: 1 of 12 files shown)


nkoenig:
@chapulina , can I bug you for a quick review please?

ros_ign_bridge/CMakeLists.txt
```diff
@@ -28,14 +28,22 @@ if ("$ENV{IGNITION_VERSION}" STREQUAL "blueprint")
   set(IGN_MSGS_VER ${ignition-msgs4_VERSION_MAJOR})
```
chapulina, line 28: Line 22 needs to be updated: Default to Dome, support Citadel and Blueprint

ros_ign_bridge/package.xml
```diff
@@ -21,8 +21,10 @@
   <!-- Default to citadel, support blueprint -->
   <depend condition="$IGNITION_VERSION == blueprint">ignition-msgs4</depend>
   <depend condition="$IGNITION_VERSION == blueprint">ignition-transport7</depend>
-  <depend condition="$IGNITION_VERSION != blueprint">ignition-msgs5</depend>
-  <depend condition="$IGNITION_VERSION != blueprint">ignition-transport8</depend>
+  <depend condition="$IGNITION_VERSION == citadel">ignition-msgs5</depend>
+  <depend condition="$IGNITION_VERSION == citadel">ignition-transport8</depend>
+  <depend condition="$IGNITION_VERSION == dome">ignition-msgs6</depend>
+  <depend condition="$IGNITION_VERSION == dome">ignition-transport9</depend>
```
chapulina, line 27: Previously, we would default to `citadel` if `IGNITION_VERSION` isn't set. I think it's important to have a default value, because users often overlook the need to set this variable.

Previously it wasn't possible to check if the variable wasn't set, that's why I checked against `!= blueprint`. But It should be possible since https://github.com/ros-infrastructure/rep/pull/288

ros_ign_gazebo/CMakeLists.txt
```diff
@@ -20,7 +20,7 @@ if ("$ENV{IGNITION_VERSION}" STREQUAL "blueprint")
   set(IGN_MSGS_VER ${ignition-msgs4_VERSION_MAJOR})
```
chapulina, line 20: Update comments on line 11

ros_ign_image/CMakeLists.txt
```diff
@@ -25,14 +25,22 @@ if ("$ENV{IGNITION_VERSION}" STREQUAL "blueprint")
   set(IGN_MSGS_VER ${ignition-msgs4_VERSION_MAJOR})
```
chapulina, line 25: Update line 19

ros_ign_image/package.xml
```diff
@@ -10,8 +10,10 @@
   <!-- Default to citadel, support blueprint -->
   <depend condition="$IGNITION_VERSION == blueprint">ignition-msgs4</depend>
   <depend condition="$IGNITION_VERSION == blueprint">ignition-transport7</depend>
-  <depend condition="$IGNITION_VERSION != blueprint">ignition-msgs5</depend>
-  <depend condition="$IGNITION_VERSION != blueprint">ignition-transport8</depend>
+  <depend condition="$IGNITION_VERSION == citadel">ignition-msgs5</depend>
+  <depend condition="$IGNITION_VERSION == citadel">ignition-transport8</depend>
+  <depend condition="$IGNITION_VERSION == dome">ignition-msgs6</depend>
+  <depend condition="$IGNITION_VERSION == dome">ignition-transport9</depend>
```
chapulina, line 16: Similar to the other comment, we need a default behavior when `IGNITION_VERSION` isn't set.

chapulina (changes requested):
Nice that it just works and no code changes are needed!

ros_ign_bridge/CMakeLists.txt
nkoenig: feff3eca0a13a4fdd841291e48385a7534fb117a

ros_ign_bridge/package.xml
nkoenig: I'm not sure I did it right  here: feff3eca0a13a4fdd841291e48385a7534fb117a

ros_ign_gazebo/CMakeLists.txt
nkoenig: feff3eca0a13a4fdd841291e48385a7534fb117a

ros_ign_image/CMakeLists.txt
nkoenig: feff3eca0a13a4fdd841291e48385a7534fb117a

ros_ign_image/package.xml
nkoenig: feff3eca0a13a4fdd841291e48385a7534fb117a

ros_ign_bridge/package.xml
chapulina: It looks like the `or` didn't work for colcon, `colcon graph` prints errors:

`ValueError: condition '$IGNITION_VERSION == dome or $IGNITION_VERSION == ''' failed to parse: Expected end of text, found 'o'`

I think we may need to have duplicate entries. These are the 2 commands that I use to check that the package files are working:

`colcon graph`

`rosdep check --from-paths src`

nkoenig: I moved empty check to separate lines in 45356162ffdd39d874867bb9c3bda554633313b8.
I don't know how to use `colcon graph`, all I get when running it is `Project +` on the command line.

chapulina: Nice! I think that worked! I updated `ros_ign_gazebo_demos` in c4b4be7.

> I don't know how to use colcon graph, all I get when running it is Project + on the command line.


Yeah I can't parse the `+` s and `*`s either. What I do to test this is to symlink my blueprint / citadel / dome workspaces into my ROS workspace and then get a dot graph. Something like:

```
cd ~/ros_ws
ln -s ~/dome_ws/src src/dome_src
colcon graph --dot | grep "ros_ign.*->.*ignition"
```

chapulina (approved):
I think this is good now, I'd just like to wait on CI to finish and to inspect it to make sure the correct dependencies are being used.

PR #128, opened 2020-12-10
Add topic flag to create robot 

### LINKED ISSUE #127: spawn robot from topic
---
Create the robot model from topic
---

Since the parameter server is no more used in ros2, the robot description is now published as a string message. Tools like rviz2 can subscribe to this message and load the robot definition. 
I wanted to get a feedback if spawning a robot given the name of the topic could be a new feature to add inside *create.cpp* in the `ros_ign_gazebo` package.

Currently the model can be loaded from different source (param, file...) and I would propose to add a new argument that let user specify the name of the topic from which to load the robot model.

It would be sufficient to run
```sh
rosrun ros_ign_gazebo create --topic /robot_description
```

I am starting to port some simulation in ros2 and ignition gazebo citadel and this feature could make my actual code cleaner, but I am not sure if maybe I am missing something.

ValerioMagnago:
Fixes #127 

## Proposed Changes

  - add new --topic parameter that enables to specify the name of the topic where to read the robot description
  - add a demo launch file that loads the robot description on the `\robot_description` topic by leveraging the `robot_state_publisher ` and then use the newly defined flag to spawn the model in gazebo and additionally the launch starts rviz2 loading the robot description
  -  check `rclcpp::ok()` while server time to download resources. In case of failure the program will return -1


ros_ign_gazebo_demos/launch/robot_description_publisher.launch.py
```diff
@@ -0,0 +1,90 @@
...
+        '<mass value="1"/>'\
+        '<inertia ixx="1" ixy="0.0" ixz="0.0" iyy="1" iyz="0.0" izz="1"/>'\
+        '</inertial>'\
+        '</link>'\
+        '</robot>" '
+
+    # Robot state publisher
+    params = {'use_sim_time': True, 'robot_description': robot_desc}
+    robot_state_publisher =  Node(
+            package='robot_state_publisher',
```
chapulina, line 53: Let's add `robot_state_publisher` as an `exec_depend` of `ros_ign_gazebo_demos` so it's installed with it

ros_ign_gazebo_demos/launch/robot_description_publisher.launch.py
```diff
@@ -0,0 +1,90 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
```
chapulina, line 1: Thanks for the demo. Mind adding an item for it to `ros_ign_gazebo_demos/README.md`?

ros_ign_gazebo/src/create.cpp
```diff
@@ -98,27 +98,49 @@ int main(int _argc, char ** _argv)
...
+      FLAGS_topic, rclcpp::QoS(1).transient_local(), fun);
+
+    rclcpp::FutureReturnCode future_ret;
+    do {
+      RCLCPP_INFO(ros2_node->get_logger(), "Waiting messages on topic [%s].", FLAGS_topic.c_str());
+      future_ret = executor.spin_until_future_complete(xml_future, timeout);
+    } while (rclcpp::ok() && future_ret != rclcpp::FutureReturnCode::SUCCESS);
+
+    if (future_ret == rclcpp::FutureReturnCode::SUCCESS) {
+      RCLCPP_INFO(ros2_node->get_logger(), xml_future.get());
```
chapulina: This could be huge depending on the model being spawned. Consider omitting this info here.

ros_ign_gazebo_demos/launch/robot_description_publisher.launch.py
```diff
@@ -0,0 +1,90 @@
...
+        '<geometry>'\
+        '<sphere radius="1.0"/>'\
+        '</geometry>'\
+        '</collision>'\
+        '<inertial>'\
+        '<mass value="1"/>'\
+        '<inertia ixx="1" ixy="0.0" ixz="0.0" iyy="1" iyz="0.0" izz="1"/>'\
+        '</inertial>'\
+        '</link>'\
+        '</robot>" '
```
chapulina: This is spawning on RViz, but failing on Gazebo. Removing this stray quote fixes it for me:

```suggestion
        '</robot>'
```

ros_ign_gazebo_demos/launch/robot_description_publisher.launch.py
```diff
@@ -0,0 +1,90 @@
...
+            name='robot_state_publisher',
+            output='screen',
+            parameters=[params],
+            arguments=[])
+
+    # Ignition gazebo
+    pkg_ros_ign_gazebo = get_package_share_directory('ros_ign_gazebo')
+    gazebo = IncludeLaunchDescription(
+        PythonLaunchDescriptionSource(
+            os.path.join(pkg_ros_ign_gazebo, 'launch', 'ign_gazebo.launch.py'))
```
chapulina: This is opening a blank world without any entities. Consider using `empty.sdf`, which has a sun and a ground plane:

```suggestion
            os.path.join(pkg_ros_ign_gazebo, 'launch', 'ign_gazebo.launch.py')),
        launch_arguments={'ign_args': '-r empty.sdf'}.items(),
```

chapulina (commented):
Thanks for the PR, it works for me. I just have some minor comments below.

CC @clalancette in case you want to take a look too.

ros_ign_gazebo_demos/launch/robot_description_publisher.launch.py
ValerioMagnago: Sure! I forgot it

ros_ign_gazebo_demos/launch/robot_description_publisher.launch.py
ValerioMagnago: Sure I will add it.
Is the name of the launch file good? Maybe you have something better in mind?

ros_ign_gazebo/src/create.cpp
ValerioMagnago: Yes you are right!

ros_ign_gazebo_demos/launch/robot_description_publisher.launch.py
chapulina: Yeah that sounds good to me, thanks!

ValerioMagnago:
@chapulina is there anything else that I can do here to help?

chapulina (approved):
Thanks for the new feature and the demo, works great for me :+1:

PR #135, opened 2021-02-23
Prevent segfault caused by missing values in sensor_msgs::JointState

peci1:
# 🦟 Bug fix

## Summary
Currently, if you send a JointState message with uneven number of items in e.g. velocity and position fields, the parameter bridge would segfault.

This PR fixes that. All missing values are substituted with NaN on the Ignition side.

This is helpful if the JointState message is used for control, or if some quantities cannot be measured for a joint (this allows to leave the whole field empty instead of manually filling NaNs/zeros in it).

## Checklist
- [ ] Added tests
- [x] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [x] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [x] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

### DIFF
--- a/ros_ign_bridge/src/convert.cpp
+++ b/ros_ign_bridge/src/convert.cpp
@@ -867,13 +867,26 @@ convert_ros_to_ign(
 {
   convert_ros_to_ign(ros_msg.header, (*ign_msg.mutable_header()));
 
-  for (auto i = 0u; i < ros_msg.position.size(); ++i)
+  const auto nan = std::numeric_limits<double>::quiet_NaN();
+  for (auto i = 0u; i < ros_msg.name.size(); ++i)
   {
     auto newJoint = ign_msg.add_joint();
     newJoint->set_name(ros_msg.name[i]);
-    newJoint->mutable_axis1()->set_position(ros_msg.position[i]);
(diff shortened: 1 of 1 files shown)


chapulina:
The [sensors_msgs/JointState docs](http://docs.ros.org/en/melodic/api/sensor_msgs/html/msg/JointState.html) say:

_All arrays in this message should have the same size, or be empty._

So I think padding with NaNs may not be the most standard approach. How about just not adding `mutable_axis1`?

peci1:
> How about just not adding `mutable_axis1`?

That's not possible. What if I want to set effort but not position? Both are a part of the same axis1. So until the position, velocity and force fields are optional in the protobuf definition, this is not possible. Because a zero position is not the same as a position that is not set (that's what the NaN expresses).

chapulina (approved):
> until the position, velocity and force fields are optional in the protobuf definition, this is not possible.

True, I didn't realize those were `double`s.

This looks good to me.

PR #142, opened 2021-03-26
[ros2] Update documentation for installation instructions and bridge examples

AndrejOrsula:
# 🦟 Bug fix

## Summary

Few minor changes to installation instructions and examples for ROS 2. Tested on Ubuntu Focal, with Edifice built from source (Docker).

### Installation instructions:
- `rosdep install` complains if Ignition libraries are included in requirements
   - Skipping keys for Citadel, Dome and Edifice versions of the required libraries (I wish there was a wildcard support)
- Both **README.md** and **ros_ign_bridge/README.md** contained installation instructions
   - Removed instructions from **ros_ign_bridge/README.md**
   - Note: Please let me know if there was a reason behind this. If so, we can revert the change and update the instructions there so that they are correct

### Examples:
- Source command of ROS 2 workspace was using invalid ROS distribution (`melodic` instead of `foxy`)
   - Fixed
- Bridge example for image transport contained a screenshot for the example using ROS 1
   - Replaced with ROS 2 alternative

## Checklist


ahcorde (approved):
I think it makes sense to just keep the install instructions in one place, the other libraries don't have install instructions in the particular README.md. 

@chapulina do you have a different opinion ?

README.md
```diff
@@ -98,23 +98,40 @@ The following steps are for Linux and OSX.
...
     git clone https://github.com/osrf/ros_ign.git -b ros2
     ```
 
-1. Install dependencies (this may also install Ignition):
+1. Install ROS dependencies:
 
     ```
     cd ~/ws
-    rosdep install --from-paths src -i -y --rosdistro foxy
+    rosdep install --from-paths src -i -y --rosdistro foxy \
```
chapulina: On Focal, rosdep should be able to install Citadel dependencies, because that's the official version for Foxy and Noetic.

After Edifice is released, that's also going to be made available through rosdep for Focal, because that will be the official version for Galactic.

Summarizing:

Ignition | Ubuntu | rosdep-installable
-- | -- | --
Citadel | Bionic | no
Citadel | Focal | yes
Dome | Bionic | no
Dome | Focal | no
Edifice | Bionic | no
Edifice | Focal | not yet

So how about, instead of skipping all keys, using `-r` to keep installing despite errors, and tell users to install Ignition from packages.osrfoundation.org if they fail through `rosdep`?

ros_ign_bridge/README.md
```diff
@@ -173,7 +124,7 @@ Now we start the ROS GUI:
 
 ```
 # Shell C:
-. /opt/ros/melodic/setup.bash
+. /opt/ros/foxy/setup.bash
 ros2 run rqt_image_view rqt_image_view /rgbd_camera/image
 ```
```
chapulina, line 130: Mind updating line 135 where it says the screenshot was taken with Kinetic?

chapulina (commented):
> Removed instructions from ros_ign_bridge/README.md

Thanks, that was already done for ROS 1 but we never updated ROS 2.

chapulina (approved):

README.md
AndrejOrsula: Good to know about Ignition `rosdep` support! Updated with `-r` (https://github.com/ignitionrobotics/ros_ign/pull/142/commits/ac729704271021d2fc309f75938cb901b76ce96c).

I wasn't really sure where to point users if `rosdep` installation fails without making it sound confusing. Especially since error(s) will be returned even if Ignition workspace overlay is sourced prior to calling `rosdep install`.

AndrejOrsula:
Note: Added one more commit into this PR to remove FluidPressure TODO from support matrix. Demo for it was enabled and tested in https://github.com/ignitionrobotics/ros_ign/pull/144, but I added the commit here to avoid merge conflicts.

PR #146, opened 2021-03-27
[ros2] Fix message duplication caused by bidirectional bridge (not for all rmw implementations)

### LINKED ISSUE #143: [ros2] Ignition Transport message duplication caused by bidirectional bridge
## Environment
* **OS Version:** Ubuntu 20.04
* **Ignition:** Edifice, built from source
* **ROS 2:** Foxy, binary install
* **ros_ign branch:** `ros2` (54e12e2653054fbb925829784ae9151c793288c3), built from source

> Note: Tested inside Docker container. I would appreciate someone to reproduce this issue to ensure it is not caused by my setup.

## Description
* **Expected behaviour:** Running a bidirectional `parameter_bridge` with Ignition publisher (separate) should generate only ROS 2 messages.
* **Actual behaviour:** Running a bidirectional `parameter_bridge` with Ignition publisher (separate) generates both ROS 2 messages and duplicate versions of ignition-transport messages.
   * **What could happening?**:
      * Bidirectional bridge creates *Node A*: Ignition subscriber + ROS 2 publisher, and *Node B*: ROS 2 subscriber + Ignition publisher.
      * Once an ignition-transport message is published, *Node A* receives it and generates ROS 2 message. Subsequently, *Node B* receives this message generated by *Node A* and generates a duplicate ignition-transport message. (problem with ROS 2 subscriber configuration?)
      * This is just a speculation, I have not investigated the code. I am sure the original developers thought of this, otherwise there would be an infinite recursion happening if this issue occurred in both directions/multiple times.
* **Workaround:** Use unidirectional bridge (99% of the use cases) to prevent message duplication that both decreases performance and could potentially break something. Updating examples to use unidirectional bridge instead of bidirectional bridge might also be an idea.

> Note: Issue does not seem to occur for ROS 2 messages, i.e. bridge does not duplicate any ROS 2 messages.
> Note: I have not tested this with ROS 1, so I am not sure if it experiences the same problem.

## Steps to reproduce

Shell 1:
`ros2 run ros_ign_bridge parameter_bridge /chatter@std_msgs/msg/String@ignition.msgs.StringMsg`

Shell 2:
`ign topic -e -t /chatter`

Shell 3:
`ign topic -t /chatter -m ignition.msgs.StringMsg -p 'data:"This message should be received only once"'`

## Output
```bash
root@P5550:~# ign topic -e -t /chatter
data: "This message should be received only once"

data: "This message should be received only once"
```

## Expected output (same as without running any bridge)
```bash
root@P5550:~# ign topic -e -t /chatter
data: "This message should be received only once"

```

AndrejOrsula:
# 🦟 Bug fix

Fixes #143 (but not for all rmw implementations)

## Summary

This PR adds [`ignore_local_publications`](https://docs.ros2.org/foxy/api/rmw/structrmw__subscription__options__t.html#ae9680be928bdaecd86fa1701304ce0a6) option to ROS 2 subscriber of the bridge, which should fix duplication of Ignition messages (echoing). **However**, [ROS 2 Design - Node to Participant mapping](https://github.com/ros2/design/blob/gh-pages/articles/170_node_to_participant_mapping.md#ignore-local-publications-option) mentions that this feature is NOT part of all rmw implementations, and unfortunately, this feature is currently not implemented in the default FastRTPS rmw of ROS 2 Foxy. When using other rmw implementations like Cyclone DDS, the issue gets fixed by this PR, otherwise it persists.

I am aware that this way of fixing the issue is not robust enough, especially as it requires non-default configuration. If you know of a better solution to this problem, please close this PR. Else the documentation probably needs to be updated a bit.

## To try out with FastRTPS (issue persists):

Shell 1:
```bash
ros2 run ros_ign_bridge parameter_bridge /chatter@std_msgs/msg/String@ignition.msgs.StringMsg
```

Shell 2:
```bash
ign topic -e -t /chatter
```

Shell 3:
```bash
ign topic -t /chatter -m ignition.msgs.StringMsg -p 'data:"This message should be received only once"'
```

## To try out with Cyclone DDS (issue is fixed):

Install Cyclone DDS
```bash
sudo apt-get update && sudo apt-get install ros-${ROS_DISTRO}-rmw-cyclonedds-cpp
```

Shell 1:
```bash
export RMW_IMPLEMENTATION=rmw_cyclonedds_cpp
ros2 run ros_ign_bridge parameter_bridge /chatter@std_msgs/msg/String@ignition.msgs.StringMsg
```

Shell 2:
```bash
ign topic -e -t /chatter
```

Shell 3:
```bash
ign topic -t /chatter -m ignition.msgs.StringMsg -p 'data:"This message should be received only once"'
```

## Checklist
- [ ] Added tests

### DIFF
--- a/ros_ign_bridge/src/factory.hpp
+++ b/ros_ign_bridge/src/factory.hpp
@@ -19,6 +19,7 @@
 
 // include ROS 2
 #include <rclcpp/rclcpp.hpp>
+#include <rclcpp/subscription_options.hpp>
 
 #include <functional>
 #include <memory>
@@ -71,9 +72,12 @@ public:
       std::placeholders::_1, ign_pub,
       ros_type_name_, ign_type_name_,
       ros_node);
+    // Ignore messages that are published from this bridge.
+    auto options = rclcpp::SubscriptionOptions();
+    options.ignore_local_publications = true;
     std::shared_ptr<rclcpp::Subscription<ROS_T>> subscription =
       ros_node->create_subscription<ROS_T>(
-      topic_name, rclcpp::QoS(rclcpp::KeepLast(queue_size)), fn);
+      topic_name, rclcpp::QoS(rclcpp::KeepLast(queue_size)), fn, options);
     return subscription;
   }
 
@@ -112,7 +116,7 @@ protected:
     ign_pub.Publish(ign_msg);
     RCLCPP_INFO_ONCE(
       ros_node->get_logger(),
-      "Passing message from ROS %s to Ignition %s (showing msg only once per type",
+      "Passing message from ROS %s to Ignition %s (showing msg only once per type)",
       ros_type_name.c_str(), ign_type_name.c_str());
   }
 


chapulina (approved):
Thanks! I think this is good to get in because it solves the problem for some RMWs. Here's an issue for support on FastRTPS: https://github.com/ros2/rmw_fastrtps/issues/76

PR #150, opened 2021-04-16
Default to Edifice for Rolling

chapulina:
# 🦟 Bug fix

## Summary

While releasing `ros_ign` for Rolling, I noticed pretty late during the release process that bloom doesn't respect local conditionals, see https://github.com/ros-infrastructure/bloom/issues/517.

I was using the `ros2` branch to release. That branch supported both Foxy and Rolling. Foxy is supposed to be released with Citadel, and Rolling (future Galactic) with Edifice. For that reason, the branch would default to Citadel if `IGNITION_VERSION` was not set. The problem is that there was no way to pass the env var to bloom to tell it to release with Edifice for Rolling.

The quickest solution right now is branching off for Rolling. I created a new `foxy` branch off the last commit before I created the changelog for the rolling release.

This PR flips the default on the `ros2` branch from Citadel to Edifice. Once this is merged, I'll rerun bloom-release.

CC @nuclearsandwich

🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,9 +16,9 @@ Eloquent | Blueprint | [dashing](https://github.com/osrf/ros_ign/tree/dashing) |
 Eloquent | Citadel | [dashing](https://github.com/osrf/ros_ign/tree/dashing) | only from source
 Eloquent | Dome | not supported |
 Foxy | Blueprint | not supported |
-Foxy | Citadel | [ros2](https://github.com/osrf/ros_ign/tree/ros2) | https://packages.ros.org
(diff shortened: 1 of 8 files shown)


j-rivero (approved):
Can not detect any error. We have that death code of pointcloud still floating around but I assume we don't want to update it without a good reason.

Issue #70, opened 2020-04-29
Bridge services


chapulina:
Ignition offers several services. It would be good to bridge them directly to ROS (1 and 2) so users can do things like spawning entities directly from ROS.

ValerioMa:
Hello @chapulina I would like to start the bridging the service since it would be useful for me to spawn a robot directly from ros2...any suggestion on how to get started?

gezp:
Hi,I have some simple and rough work about it:  https://github.com/gezp/ros_ign_gazebo_py

it's very useful to control ignition Gazebo (by service) with ROS , for example, resume and pasue Ignition Gazebo, create model,delete model and set pose of model. 

In my case, I need control Ignition Gazebo with python, but `ign-transport` doesn't support python, so i use ROS as bridge with C++, and then i create ROS Client Node with python so that i can control ignition Gazebo with python.

by the way, before implementing bridge services,  `ros_ign_interfaces` is necessary to provide ros msg,srv files related to `ign-msgs` , like `EntityFactory.msg` . i think this package should be considered currently, and this is my opinion, any discussions is welcome.

finally, this is a very useful function, i hope it can be implemented earlier.

chapulina:
Hi @gezp , we're not currently planning to work on this, but we can review a pull request if you're willing to implement it.

gezp:
I'm glad to contribute, but the New ROS Package `ros_ign_interfaces`  is allowed to pull request? 

In addition, I think `ign_gazebo_manager.cpp` (https://github.com/gezp/ros_ign_gazebo_py : bridge part services ) should be put in `ros_ign_gazebo/src`  , and `ign_gazebo_interface.py` (ros client node for ign_gazebo_manager service, provide python interface )should be put in `ros_ign_gazebo/ros_ign_gazebo` (python package will be created).

 I will do this work if it's ok.

chapulina:
> New ROS Package ros_ign_interfaces is allowed to pull request?

Yes, thanks! I think there's no way around it, we'll need to add some Ignition-specific ROS messages that aren't provided by other packages.

I considered maybe using [gazebo_msgs](https://github.com/ros-simulation/gazebo_ros_pkgs/tree/kinetic-devel/gazebo_msgs), but those messages carry some Gazebo-classic concepts that I think we wouldn't want here.

>  I think ign_gazebo_manager.cpp (https://github.com/gezp/ros_ign_gazebo_py : bridge part services ) should be put in ros_ign_gazebo/src

[...]

PR #152, opened 2021-05-04
[ros2]  new package ros_ign_interfaces, provide some  Ignition-specific ROS messages.

gezp:
This is migration of [gazebo_ros_pkgs/gazebo_msgs](https://github.com/ros-simulation/gazebo_ros_pkgs/tree/foxy/gazebo_msgs) for Ignition to provide some  Ignition-specific ROS messages. 

This package is necessary to implement some functions like this #70 .

* `ContactState.msg` and `ContactStates.msg` are migrated from [gazebo_ros_pkgs/gazebo_msgs](https://github.com/ros-simulation/gazebo_ros_pkgs/tree/foxy/gazebo_msgs) 
* `DeleteEntity.srv` and `SpawnEntity.srv` are migrated from [gazebo_ros_pkgs/gazebo_msgs](https://github.com/ros-simulation/gazebo_ros_pkgs/tree/foxy/gazebo_msgs)  and modified.
*  `ControlWorld.srv`  refer to [ign-msgs/proto/ignition/msgs/world_control.proto](https://github.com/ignitionrobotics/ign-msgs/blob/ign-msgs7/proto/ignition/msgs/world_control.proto)
*  `SetEntityPose.srv`  is new srv file.


ros_ign_interfaces/package.xml
```diff
@@ -0,0 +1,25 @@
+<package format="3">
+  <name>ros_ign_interfaces</name>
+  <version>0.233.1</version>
+  <description>Message and service data structures for interacting with Ignition from ROS2.</description>
+  <license>Apache 2.0</license>
+  <maintainer email="louise@openrobotics.org">Louise Poubel</maintainer>
+
```
chapulina: Feel free to add yourself as author :wink:

ros_ign_interfaces/msg/ContactState.msg
```diff
@@ -0,0 +1,8 @@
+string info                                   # text info on this contact
```
chapulina: I know the message on `gazebo_msgs` has the info field, which is filled with debug information, but I believe this isn't a good practice. It makes the message larger and doesn't add any new information.

How about removing the field?

ros_ign_interfaces/srv/ControlWorld.srv
```diff
@@ -0,0 +1,11 @@
+bool pause
```
chapulina: Do you mind documenting all the fields in the message? Let's try to do better than `ign-msgs` :grimacing:

chapulina: Do you mind calling this message `WorldControl` to match [ignition::msgs::WorldControl](https://github.com/ignitionrobotics/ign-msgs/blob/ign-msgs7/proto/ignition/msgs/world_control.proto)?

ros_ign_interfaces/srv/DeleteEntity.srv
```diff
@@ -0,0 +1,5 @@
+string name                       # Name of the Ignition Gazebo entity to be deleted. 
+uint8 type                        # Type: Model or Light
```
chapulina: How about adding an entity message like [ignition::msgs::Entity](https://github.com/ignitionrobotics/ign-msgs/blob/ign-msgs7/proto/ignition/msgs/entity.proto) and using it here? Using the same pattern as Ignition should make it easier to maintain. It's also easier to only update one messsage if we add new entity types.

ros_ign_interfaces/msg/ContactState.msg
```diff
@@ -0,0 +1,8 @@
...
+string collision1_name                        # name of contact collision1
+string collision2_name                        # name of contact collision2
+geometry_msgs/Wrench[] wrenches               # list of forces/torques
+geometry_msgs/Wrench total_wrench             # sum of forces/torques in every DOF
```
chapulina: I'd also suggest removing the `total_wrench` field, because it doesn't add new information. The receiver can calculate it if they need it.

ros_ign_interfaces/msg/ContactState.msg
```diff
@@ -0,0 +1,8 @@
...
+geometry_msgs/Vector3[] contact_positions     # list of contact position
+geometry_msgs/Vector3[] contact_normals       # list of contact normals
```
chapulina: Prefixing these fields with `contact_` feels redundant.

ros_ign_interfaces/msg/ContactState.msg
```diff
@@ -0,0 +1,8 @@
...
+float64[] depths                              # list of penetration depths
```
chapulina: I'd suggest changing the name of this message to `Contact` so it matches [ignition::msgs::Contact](https://github.com/ignitionrobotics/ign-msgs/blob/ign-msgs7/proto/ignition/msgs/contact.proto).

ros_ign_interfaces/srv/SpawnEntity.srv
```diff
@@ -0,0 +1,12 @@
+string name                             # New name for the entity, overrides the name on the SDF
```
chapulina: How about renaming this message to `EntityFactory` so it matches `ign-msgs`?

chapulina (commented):
Thank you for the PR! I think this is an opportunity to improve on the message definitions from `gazebo_msgs`, otherwise we could just depend on that package.

Some thoughts:

* The messages should match [ign-msgs](https://github.com/ignitionrobotics/ign-msgs/tree/ign-msgs7/proto/ignition/msgs) as much as possible, for seamless integration. This includes matching message names and fields.
* We should remove textual debug fields like `info` and `status_message`, which aren't a standard practice on ROS and just make the messages larger. This kind of information can go into console logging instead. It's also difficult to standardize textual status messages so the receiver knows what to expect.

gezp:
According to above advice, I modified and added the definition of message to make them match `ign-msgs` .

gezp:
@chapulina 
hi, i have modified my PR, could you review again? after this PR completed, I could try to implement service bridge.

ros_ign_interfaces/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+endif()
+
+
+find_package(ament_cmake REQUIRED)
+find_package(builtin_interfaces REQUIRED)
+find_package(std_msgs REQUIRED)
+find_package(geometry_msgs REQUIRED)
+find_package(rosidl_default_generators REQUIRED)
+
+
```
ahcorde: nit:
```suggestion
```

ros_ign_interfaces/README.md
```diff
@@ -0,0 +1,20 @@
+# Message and service data structures for interacting with Ignition from ROS2
+
+This package currently contains some Ignition-specific ROS message and service data structures (.msg and .srv)
+
+## Messages (.msg)
+
+* [Contact](msg/Contact.msg) : related to [ignition::msgs::Contact](https://github.com/ignitionrobotics/ign-msgs/blob/ign-msgs7/proto/ignition/msgs/contact.proto). Contant info bewteen collisions in Ignition Gazebo.
```
ahcorde: Same below

```suggestion
* [Contact](msg/Contact.msg): related to [ignition::msgs::Contact](https://github.com/ignitionrobotics/ign-msgs/blob/ign-msgs7/proto/ignition/msgs/contact.proto). Contant info bewteen collisions in Ignition Gazebo.
```

ros_ign_interfaces/msg/Entity.msg
```diff
@@ -0,0 +1,13 @@
+#Entity type: constant definition
+uint8 NONE = 0
```
ahcorde: can you indent the default values at the same level ?

ros_ign_interfaces/msg/Entity.msg
```diff
@@ -0,0 +1,13 @@
...
+uint8 MODEL = 2
+uint8 LINK = 3
+uint8 VISUAL = 4
+uint8 COLLISION = 5
+uint8 SENSOR = 6
+uint8 JOINT = 7
+
+uint64 id      #Entity unique identifier accross all types. Defaults to 0
+string name    #Entity name, which is not guaranteed to be unique.
+uint8 type     #Entity type.
```
ahcorde:
```suggestion
uint64 id      # Entity unique identifier accross all types. Defaults to 0
string name    # Entity name, which is not guaranteed to be unique.
uint8 type     # Entity type.
```

ros_ign_interfaces/msg/WorldControl.msg
```diff
@@ -0,0 +1,8 @@
+bool pause                                  # paused state.
```
ahcorde: Start comment with a Upper case ( in all msgs )
```suggestion
bool pause                                  # Paused state.      
```

ahcorde (changes requested):
Some style nits

ros_ign_interfaces/package.xml
```diff
@@ -0,0 +1,26 @@
+<package format="3">
+  <name>ros_ign_interfaces</name>
+  <version>0.233.1</version>
```
ahcorde, line 3: Last comment

can you change here the version to match with the others ?
```suggestion
  <version>0.9.5</version>
```

ahcorde (changes requested):

ros_ign_interfaces/package.xml
gezp: the version of packages  in ros2 branch  is 0.233.1, and currently this package is created for ros2.

ahcorde (approved):

PR #156, opened 2021-05-27
Joint states tutorial

vatanaksoytezer:
Adds an rrbot model to demos and shows the usage of joint_states plugin.


ros_ign_gazebo_demos/launch/joint_states.launch.py
```diff
@@ -0,0 +1,89 @@
...
+            os.path.join(pkg_ros_ign_gazebo, 'launch', 'ign_gazebo.launch.py')),
+            launch_arguments={'ign_args': '-r empty.sdf'}.items(),
+    )
+
+    # RViz
+    rviz = Node(
+        package='rviz2',
+        executable='rviz2',
+        arguments=['-d', os.path.join(pkg_ros_ign_gazebo_demos, 'rviz', 'joint_states.rviz')],
+        parameters=[]
```
ahcorde: remove

ros_ign_gazebo_demos/launch/joint_states.launch.py
```diff
@@ -0,0 +1,89 @@
...
+    )
+
+    # Spawn
+    rrbot_sdf_path =  os.path.join(pkg_ros_ign_gazebo_demos, 'models', 'rrbot.sdf')
+    # rrbot_sdf_path =  os.path.join(pkg_ros_ign_gazebo_demos, 'models', 'rrbot2.sdf')
```
ahcorde: remove

ros_ign_gazebo_demos/launch/joint_states.launch.py
```diff
@@ -0,0 +1,89 @@
...
+    print(rrbot_sdf_path)
```
ahcorde: remove

ros_ign_gazebo_demos/launch/joint_states.launch.py
```diff
@@ -0,0 +1,89 @@
...
+    spawn = Node(package='ros_ign_gazebo', executable='create',
+                arguments=[
+                    '-name', 'rrbot',
+                    '-file', rrbot_sdf_path,
+                    ],
+                output='screen',
+                )
```
ahcorde, line 52: parentheses at the same indentation as the others

ros_ign_gazebo_demos/launch/joint_states.launch.py
```diff
@@ -0,0 +1,89 @@
+import os
+import xacro
```
ahcorde, line 2: you are importing here xacro, but you are not using it

ahcorde (changes requested):
You are adding here the URDF and the SDF and using it both in the launch file. I don't like this approach because if you modify one you must need to modify the other one. You should use only the URDF, with xacro (which is imported but not used) you can convert the URDF and uing it to spawn the model. In this [file](https://github.com/ros-simulation/gazebo_ros_demos/blob/a5b8877fc8398e044e4aedccf69afb6c39ee298b/rrbot_gazebo/launch/rrbot_world.launch.py) you can find how to do it.

Another question that I open here it's that the `rrbot.urdf` should live only in one repository we should not copy this file in every repo that make use of it.

ahcorde (changes requested):
You should also sign the [PR](https://github.com/ignitionrobotics/ros_ign/pull/156/checks?check_run_id=2685739378)

ahcorde:
@vatanaksoytezer, I have seen in your profile that you are working for Picknik maybe we can speed up this other PR https://github.com/ros-simulation/gazebo_ros_demos/pull/41 where I think the rrobot description should live.

vatanaksoytezer:
> @vatanaksoytezer, I have seen in your profile that you are working for Picknik maybe we can speed up this other PR [ros-simulation/gazebo_ros_demos#41](https://github.com/ros-simulation/gazebo_ros_demos/pull/41) where I think the rrobot description should live.

Sure I will ping Dave internally.

ros_ign_gazebo_demos/launch/joint_states.launch.py
vatanaksoytezer: Oops forgot those in cleanup

vatanaksoytezer:
> Another question that I open here it's that the rrbot.urdf should live only in one repository we should not copy this file in every repo that make use of it.

I would agree that but I really wouldn't want to make `ign_gazebo_demos` depend to `gazebo_ros_demos` for rrbot only. Maybe you guys can consider opening up a rrbot description and depend on that?

vatanaksoytezer:
> You are adding here the URDF and the SDF and using it both in the launch file. I don't like this approach because if you modify one you must need to modify the other one. You should use only the URDF, with xacro (which is imported but not used) you can convert the URDF and uing it to spawn the model. In this [file](https://github.com/ros-simulation/gazebo_ros_demos/blob/a5b8877fc8398e044e4aedccf69afb6c39ee298b/rrbot_gazebo/launch/rrbot_world.launch.py) you can find how to do it.

@ahcorde Follow up question on this one. Since this is to demonstrate the usage of `joint_states` plugin in ignition, where should I put the plugin if I don't have the sdf file? I was mostly looking up for @chapulina 's `dolly` package where she also defines the robot from  an sdf file.

vatanaksoytezer:
Another problem that I just realised is if rrbot_description live somewhere else and contains `gazebo` and `ros2_control` stuff that is not compatible with ignition (yet), wouldn't it be problematic?

vatanaksoytezer:
@ahcorde update: 

- Pinged Dave on gazebo PR.
- Signed off the commits and cleaned up stuff, except for the sdf (see my comments above)

vatanaksoytezer:
Ignore the comments above on sdf, just figured that out. I'll fix in and send updates soon

vatanaksoytezer:
This should work, but the colors (and I think frictions) are not parsed correctly to ign. If you point me a way I can fix it right away

ahcorde:
@vatanaksoytezer, I can see the color in ignition 

Selección_104

ahcorde (approved):
LGTM

PR #185, opened 2021-10-28
Default to Fortress

doisyg:
Now that Fortress is out, it makes sense to target it by default (+ it is a LTS).

### DIFF
--- a/ros_ign_bridge/CMakeLists.txt
+++ b/ros_ign_bridge/CMakeLists.txt
@@ -39,17 +39,8 @@ elseif("$ENV{IGNITION_VERSION}" STREQUAL "dome")
   set(IGN_MSGS_VER ${ignition-msgs6_VERSION_MAJOR})
 
   message(STATUS "Compiling against Ignition Dome")
-# Fortress
+# Edifice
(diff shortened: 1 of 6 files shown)


chapulina (changes requested):
Thanks for the PR!

The default Ignition version for each ROS release is defined on [REP-2000](https://www.ros.org/reps/rep-2000.html#galactic-geochelone-may-2021-november-2022). The `ros2` branch is currently used for both Galactic and Rolling. The plan is to use Fortress with Humble, and we're working on upgrading to Fortress on Rolling asap, as mentioned during [yesterday's community meeting](https://www.youtube.com/watch?v=CY-_Yvu0ya0). 

But before merging this PR, there are many steps to be done first (captured in https://github.com/ignition-tooling/release-tools/issues/459). We'll also need to create a new `galactic` branch that will continue to default to Edifice (more context in https://github.com/ignitionrobotics/ros_ign/pull/150).

I'll mark with "request for changes" to make sure we don't merge this too soon.

chapulina:
Thanks for kicking this off @doisyg . Your commit was merged as part of #195

PR #191, opened 2021-12-21
Break tus

mjcarroll:
Draft of breaking up translation units to help with memory/CPU usage of compilation.


ros_ign_bridge/src/factories/visualization_msgs.hpp
```diff
@@ -0,0 +1,36 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef ROS_IGN_BRIDGE__FACTORIES__VISUALIZATION_MSGS_HPP_
+#define ROS_IGN_BRIDGE__FACTORIES__VISUALIZATION_MSGS_HPP_
+
+// ROS messages
+
+// Ignition messages
+#include <ignition/msgs.hh>
```
chapulina, line 21: Another thing that may help compilation times is to use more targeted includes, i.e. the specific `pb.h` instead of `msgs.hh`.

In this specific case, I think we can even remove this include though

ros_ign_bridge/CMakeLists.txt
```diff
@@ -60,11 +60,36 @@ endif()
...
+  src/factories/geometry_msgs.cpp
+  src/factories/nav_msgs.cpp
+  src/factories/rosgraph_msgs.cpp
+  src/factories/sensor_msgs.cpp
+  src/factories/std_msgs.cpp
+  src/factories/visualization_msgs.cpp
+)
+
+set(bridge_lib
+  ros_ign_bridge_lib
```
chapulina, line 83: CI isn't happy

```
CMake Error at CMakeLists.txt:99 (add_library):
  add_library cannot create target "ros_ign_bridge_lib" because another
  target with the same name already exists.  The existing target is a shared
  library created in source directory
```

chapulina (commented):
Nice! Thanks for working on this :bowing_woman: I think we aren't having issues with compilation memory for ROS 1 though, so we'll only know if this works when we port it to the `ros2` branch.

mjcarroll:
> Nice! Thanks for working on this bowing_woman I think we aren't having issues with compilation memory for ROS 1 
though, so we'll only know if this works when we port it to the ros2 branch.

This was left over from ROS World prep and I found it when cleaning out some older workspaces.  I can try to port to ROS 2 if I get some time, I just wanted to make sure it didn't get lost.

Issue #190, opened 2021-12-16
ros_ign_gazebo create.cpp usage message incorrect


oldtopos:
and misleading

## Environment
* OS Version: Ubuntu 20.04
* Source build, ros2 branch

## Description
* Expected behavior: 
* 
* gflags::SetUsageMessage(
    R"(Usage: create -world [arg] [-file FILE] [-param PARAM] [-topic TOPIC]
                       [-string STRING] [-name NAME] [-x X] [-y Y] [-z Z]
                       [-R ROLL] [-P PITCH] [-Y YAW])");
* Actual behavior:   
* 
* gflags::SetUsageMessage(
    R"(Usage: create -world [arg] [-file FILE] [-param PARAM] [-topic TOPIC]
                       [-string STRING] [-name NAME] [-X X] [-Y Y] [-Z Z]
                       [-Roll ROLL] [-Pitch PITCH] [-Yaw YAW])");

chapulina:
Oops, that's here:

https://github.com/ignitionrobotics/ros_ign/blob/5d9cfef3d58948a77b962d231e281c88ea88e188/ros_ign_gazebo/src/create.cpp#L51

Aryaman22102002:
I have tried to fix this issue. Can someone please take a look at the pull request.  Link to pr: https://github.com/ignitionrobotics/ros_ign/pull/206

Aryaman22102002:
Does this issue needs to be resolved for other branches too?

chapulina:
> Does this issue needs to be resolved for other branches too?

Yeah I think so. If you could open a PR for `foxy`, we can then forward-port it to `galactic`. Thanks!

chapulina:
Thanks @Aryaman22102002

PR #206, opened 2022-01-12
create.cpp usage message corrected

Aryaman22102002:
This fixes the issue  ros_ign_gazebo create.cpp usage message incorrect #190 .


ahcorde (changes requested):
Option `[-topic TOPIC]` is not available in this branch

https://github.com/ignitionrobotics/ros_ign/blob/af128814e3e615a8f43d8a7b13fefa24e8a3048c/ros_ign_gazebo/src/create.cpp#L26-L37

is there something that I'm missing ?

Aryaman22102002 (commented):
[-topic TOPIC] added in create.cpp

PR #207, opened 2022-01-12
create.cpp usage message fixed for ros2 branch

Aryaman22102002:
Signed-off-by: Aryaman <aryashardul2002@gmail.com>

This fixes  ros_ign_gazebo create.cpp usage message incorrect #190 .


chapulina (commented):
Thanks for the fix! I just noticed that `allow_renaming` is also missing, mind adding it too?

Aryaman22102002 (commented):
allow_renaming has now been added in usage message.

chapulina (approved):
Thanks!

PR #212, opened 2022-01-27
Break apart ros_subscriber test translation unit

mjcarroll:
Another attempt at reducing compilation memory usage.

Previously, building `ros_subscriber.cpp` used about 1.1 GB of RAM according to `templight`.

After splitting the individual TUs use:

image


ros_ign_bridge/test/subscribers/ros_subscriber/geometry_msgs_subscriber.cpp
```diff
@@ -0,0 +1,145 @@
...
+// limitations under the License.
+//
+
+#include <gtest/gtest.h>
+
+#include <chrono>
+#include <memory>
+#include <string>
+
+#include <rclcpp/rclcpp.hpp>
```
chapulina: Do we need `rclcpp` on these cpp files?

ros_ign_bridge/test/subscribers/ros_subscriber/nav_msgs_subscriber.cpp
```diff
@@ -0,0 +1,37 @@
...
+// limitations under the License.
+//
+
+#include <gtest/gtest.h>
+
+#include <chrono>
+#include <memory>
+#include <string>
+
+#include <rclcpp/rclcpp.hpp>
```
chapulina: I think we can remove `rclcpp` and `string`

chapulina (approved):
LGTM, I think we can get rid of some includes too for extra optimization

ros_ign_bridge/test/subscribers/ros_subscriber/geometry_msgs_subscriber.cpp
mjcarroll: 49be345

ros_ign_bridge/test/subscribers/ros_subscriber/nav_msgs_subscriber.cpp
mjcarroll: 49be345

PR #213, opened 2022-01-28
Bring ros2 branch up-to-date with Rolling

mjcarroll:
In order to build clean on rolling:

* change the deprecated use of `declare_parameters`
* Add a pure virtual destructor to FactoryInterface (at least for clang/llvm)
* Make linters pass

Broke this off of #212 for easier review

### DIFF
--- a/ros_ign_bridge/src/factory_interface.hpp
+++ b/ros_ign_bridge/src/factory_interface.hpp
@@ -15,14 +15,14 @@
 #ifndef  FACTORY_INTERFACE_HPP_
 #define  FACTORY_INTERFACE_HPP_
 
-// include ROS 2
-#include <rclcpp/rclcpp.hpp>
-
 // include Ignition Transport
 #include <ignition/transport/Node.hh>
 
 #include <memory>
 #include <string>
+
+// include ROS 2
+#include <rclcpp/rclcpp.hpp>
 
 namespace ros_ign_bridge
 {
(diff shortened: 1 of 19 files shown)


ros_ign_bridge/src/factory_interface.hpp
```diff
@@ -15,21 +15,24 @@
...
+#include <rclcpp/rclcpp.hpp>
+
 namespace ros_ign_bridge
 {
 
 class FactoryInterface
 {
 public:
+  virtual
+  ~FactoryInterface() = 0;
```
chapulina, line 34:
```suggestion
  ~FactoryInterface();
```

It's not really pure because we're providing the implementation, right?

chapulina (approved):
LGTM, I just have a question about the pure-virtual destructor

ros_ign_bridge/src/factory_interface.hpp
mjcarroll: You still have to provide a body for a pure-virtual destructor.  There is probably a more official resource, but this confirms: https://www.geeksforgeeks.org/pure-virtual-destructor-c/

chapulina: Oh gosh, ok. Interestingly, before posting the suggestion I tried it locally and it compiles fine (at least with gcc) :woman_shrugging:

mjcarroll: The only reason I caught it was because of templight, which uses clang under the hood.

PR #215, opened 2022-02-07
[galactic] backport test memory usage improvements

mjcarroll:
Backport two test CPU/memory improvements to get Galactic debbuild to turn over.


cottsay (approved):
Looks like a clean backport, and the builds came back green. Thanks for fixing this.

I wonder of [precompiled headers](https://cmake.org/cmake/help/v3.20/command/target_precompile_headers.html) could also be used to optimize this process? It's also possible that it would make memory usage worse - I have little experience with PCH.

mjcarroll:
>I wonder of [precompiled headers](https://cmake.org/cmake/help/v3.20/command/target_precompile_headers.html) could also be used to optimize this process? It's also possible that it would make memory usage worse - I have little experience with PCH.

I think we could try it.  I seem to recall there being cross-platform issues with PCH in the past, so I haven't looked at them recently.

PR #211, opened 2022-01-19
Support bridging services

ivanpauno:
# 🎉 New feature

Closes #70

## Summary

* Add support for bridging services.
  At the moment you can only bridge the world control service.
* Add some message definitions that can be useful for bridging other ignition services.

I plan to clean up the code a bit and support to bridge more services in a follow up PR.

## Test it

```bash
ros ign /path/to/any/world/with/ign/gui/world/control/enabled
ros2 run ros_ign_bridge parameter_bridge /world/wheel_slip/control@ros_ign_interfaces/srv/ControlWorld
ros2 service call /world/<world_name>/control ros_ign_interfaces/srv/ControlWorld "world_control:
  pause: false
  step: false
  multi_step: 0
  reset:
    all: false
    time_only: false
    model_only: false
  seed: 0
  run_to_sim_time:
    sec: 0
    nanosec: 0"
```

## Checklist
- [ ] Signed all commits for DCO
- [ ] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #216, opened 2022-02-08
Add conversions for ros_ign_interfaces/WorldControl and builtin_interfaces/Time

ivanpauno:
# 🎉 New feature

Split from #211.

## Summary

Add the following conversions:

* ros_ign_interfaces/WorldControl <-> ignition.msgs.WorldControl
* builtin_interfaces/Time <-> ignition.msgs.Time

## Checklist
- [x] Added tests
- [x] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [x] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))

### DIFF
--- a/ros_ign_bridge/CMakeLists.txt
+++ b/ros_ign_bridge/CMakeLists.txt
@@ -11,6 +11,7 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
 endif()
 
 find_package(ament_cmake REQUIRED)
+find_package(builtin_interfaces REQUIRED)
 find_package(geometry_msgs REQUIRED)
 find_package(nav_msgs REQUIRED)
 find_package(rclcpp REQUIRED)
@@ -42,13 +43,14 @@ else()
 endif()
 
 set(PACKAGES
-  std_msgs
+  builtin_interfaces
   geometry_msgs
   nav_msgs
-  tf2_msgs
   ros_ign_interfaces
   rosgraph_msgs
   sensor_msgs
+  std_msgs
+  tf2_msgs
   trajectory_msgs
 )
 
@@ -84,6 +86,8 @@ foreach(PKG ${PACKAGES})
     PROPERTIES POSITION_INDEPENDENT_CODE ON)
 
 endforeach()
+target_link_libraries(ros_ign_interfaces_bridge builtin_interfaces_bridge)
+target_link_libraries(std_msgs_bridge builtin_interfaces_bridge)
 
 set(bridge_executables
   parameter_bridge
@@ -162,6 +166,7 @@ if(BUILD_TESTING)
     ignition-transport${IGN_TRANSPORT_VER}::core
   )
   ament_target_dependencies(test_utils
+    builtin_interfaces
     geometry_msgs
     nav_msgs
     rclcpp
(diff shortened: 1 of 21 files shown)


ros_ign_bridge/README.md
```diff
@@ -37,6 +38,7 @@ service calls. Its support is limited to only the following message types:
 | ros_ign_interfaces/msg/StringVec     | ignition::msgs::StringMsg_V          |
 | ros_ign_interfaces/msg/TrackVisual   | ignition::msgs::TrackVisual          |
 | ros_ign_interfaces/msg/VideoRecord   | ignition::msgs::VideoRecord          |
+| ros_ign_interfaces/msg/WorldControl  | ignition::msgs::WorldContro          |
```
mjcarroll:
```suggestion
| ros_ign_interfaces/msg/WorldControl  | ignition::msgs::WorldControl         |
```

ros_ign_bridge/CMakeLists.txt
```diff
@@ -84,6 +86,8 @@ foreach(PKG ${PACKAGES})
     PROPERTIES POSITION_INDEPENDENT_CODE ON)
 
 endforeach()
+target_link_libraries(ros_ign_interfaces_bridge builtin_interfaces_bridge)
```
mjcarroll: I wonder if it makes sense to include `builtin_interfaces` in the `bridge_utils` library above, since almost everything else is going to use it?

ivanpauno: There are a lot of interdependencies between interface packages, so I don't think adding `builtin_interfaces` to `bridge_utils` will help a lot.

e.g.: we didn't need to add a similar dependency (ros_ign_interfaces_bridge -> geometry_msgs_bridge) for this [line](https://github.com/ignitionrobotics/ros_ign/blob/6334294aac55a02158eb56d2f1505997db43fc55/ros_ign_bridge/src/convert/ros_ign_interfaces.cpp#L156). For some reason it works without it.

Why are we generating a `*_bridge` static library for each interface package?
We're only static linking them once into the dynamic library, so just adding all the source files to the dynamic library will result in cmake generating the same steps (i.e. if we're not using the static library more than once, there's no benefit of having them).

mjcarroll: > Why are we generating a *_bridge static library for each interface package?

It was a way to force compilation of each TU separately.  We were running up against memory/cpu limits any time another message definition was added in the deb builder stage.  We may be able to drop the static libraries at this point, but we needed to get it building for galactic/rolling syncs unblocked.

ivanpauno: I think those lines weren't needed, so removing them in https://github.com/ignitionrobotics/ros_ign/pull/216/commits/d0b6d99870a3ebfa71fbae1e62ff25d4a097065c.

---

> It was a way to force compilation of each TU separately

That's already the case when `add_library()` is used AFAIK. If an add_library() call has multiple source file, cmake will generate an object file for each one and then link everything together (checked one of the Makefiles cmake generates, that seems to be the case).
Maybe there's a subtle difference between both ways, I'm not sure.

If CI was running out of memory, I think we probably can modularize some of the header files that include ALL message files to avoid the issue.

mjcarroll: > Maybe there's a subtle difference between both ways, I'm not sure.

Probably not, this was probably a bit aggressive to try to clear the buildfarm.  I'll make a note to re-evaluate now that things are quieter and there isn't any sync pressure.

mjcarroll (approved):

PR #232, opened 2022-04-04
Factory interface needs virtual destructor

mjcarroll:
Suppresses some warnings with clang-12.


mjcarroll:
`Gpr` is only red because of 

```
Policy CMP0072 is not set: FindOpenGL prefers GLVND by default when
```

chapulina (approved):

PR #233, opened 2022-04-04
Generate boilerplate files from Python scripts

mjcarroll:
The way that we add factories can be a bit error-prone, as there are a lot of strings that cannot be checked at compilation time.  This changes several of the boilerplate files to be generated automatically by python scripts, in line with how `ros1_bridge` does it.

For ease of review, I have broken this into a few commits.

* Adds the python script
*  Adds support headers and CMake for the generated files
*  Remove files that are now generated
*  Add a helper utility for messages introspection.

Now to add a message mapping, we can update this `MAPPING` dictionary : `ros_ign_bridge/ros_ign_bridge/mappings.py`, as well as making sure that the appropriate functions are implemented for the actual conversions.

Some other helpful additions:

* `ros_ign_bridge/bin/ros_ign_bridge_markdown_table` will allow you to generate the markdown table from the README.md given the currently available mappings.
* `bridge_types` gives an ability to introspect into types that are compiled in.  Could potentially be used with runtime checks via `launch` to make sure that correct bridges are used:
  * `ros2 run ros_ign_bridge bridge_types --print-pairs` - Print known mapping pairs
 ```
Supported ROS <=> Ignition  message type conversion pairs:
  - 'geometry_msgs/msg/Point' (ROS) <=> 'ignition.msgs.Vector3d' (IGN)
  - 'geometry_msgs/msg/Pose' (ROS) <=> 'ignition.msgs.Pose' (IGN)
  - 'geometry_msgs/msg/PoseStamped' (ROS) <=> 'ignition.msgs.Pose' (IGN)
...
  - 'trajectory_msgs/msg/JointTrajectory' (ROS) <=> 'ignition.msgs.JointTrajectory' (IGN)
 ```
  * `ros2 run ros_ign_bridge bridge_types --print-ign` - Print known valid Ignition types
  * `ros2 run ros_ign_bridge bridge_types --print-ros` - Print known valid ROS types
  * `ros2 run ros_ign_bridge bridge_types --get-ros-to-ign` - Get correct ignition type from ros type, return 1 if arg isn't a valid ros type
```
ros2 run ros_ign_bridge bridge_types --get-ros-to-ign std_msgs/msg/UInt32
ignition.msgs.UInt32
```
  * `ros2 run ros_ign_bridge bridge_types --get-ign-to-ros` - Get correct ROS type from Ignition type, return 1 if it's not a valid ignition type
```
ros2 run ros_ign_bridge bridge_types --get-ign-to-ros ignition.msgs.UInt32
std_msgs/msg/UInt32
```


ros_ign_bridge/src/bridge_types.cpp
```diff
@@ -0,0 +1,113 @@
...
+  } else if (get_flag_option(args, "--get-ign-to-ros")) {
+    if (args.size() < 3) {
+      printf("--get-ign-to-ros requires an argument\n");
+      return 1;
+    }
+
+    std::string ros_type_name;
+    auto found = ros_ign_bridge::get_ign_to_ros_mapping(args[2], ros_type_name);
+    if (found) {
+      printf("%s\n", ros_type_name.c_str());
```
ahcorde: why do you use `printf`, you included `iostream` ?

ros_ign_bridge/src/bridge_types.cpp
```diff
@@ -0,0 +1,113 @@
+// Copyright 2022 Open Source Robotics Foundation, Inc.
```
ahcorde, line 1: when I run 
```bash
ros2 run ros_ign_bridge bridge_types --get-ign-to-ros ignition.msgs.UI
```

There is no output because the type doesn't exists, but I don't know if something happens in the background. Can you add a message? maybe something like "This type is not available use the option --print-ign to show the available messages"

ros_ign_bridge/bin/ros_ign_bridge_markdown_table
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python3
```
ahcorde, line 1: License

ahcorde (changes requested):

ros_ign_bridge/src/bridge_types.cpp
mjcarroll: It's what was used where I copied from: https://github.com/ros2/ros1_bridge/blob/master/src/dynamic_bridge.cpp#L95-L115

I will update to use iostream.

ros_ign_bridge/src/bridge_types.cpp
mjcarroll: It returns non-zero, which is the most helpful for scripting, but I can add a human-readable message as well.

ahcorde (approved):

PR #239, opened 2022-04-08
Add Dataframe message and bridging

mjcarroll:

### DIFF
--- a/ros_ign_bridge/include/ros_ign_bridge/convert/ros_ign_interfaces.hpp
+++ b/ros_ign_bridge/include/ros_ign_bridge/convert/ros_ign_interfaces.hpp
@@ -19,6 +19,7 @@
 #include <ros_ign_interfaces/msg/joint_wrench.hpp>
 #include <ros_ign_interfaces/msg/contact.hpp>
 #include <ros_ign_interfaces/msg/contacts.hpp>
+#include <ros_ign_interfaces/msg/dataframe.hpp>
 #include <ros_ign_interfaces/msg/light.hpp>
 
 // Ignition messages
@@ -26,6 +27,7 @@
 #include <ignition/msgs/joint_wrench.pb.h>
 #include <ignition/msgs/contact.pb.h>
 #include <ignition/msgs/contacts.pb.h>
+#include <ignition/msgs/dataframe.pb.h>
 #include <ignition/msgs/light.pb.h>
 
 #include <ros_ign_bridge/convert_decl.hpp>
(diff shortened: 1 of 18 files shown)


mjcarroll:
CI may fail without the updated ign-msgs package here

caguero (approved):
Works for me!

iche033:
do we need to set a min required ign-msgs8 version here?
https://github.com/ignitionrobotics/ros_ign/blob/galactic/ros_ign_bridge/CMakeLists.txt#L29

caguero:
> do we need to set a min required ign-msgs8 version here? https://github.com/ignitionrobotics/ros_ign/blob/galactic/ros_ign_bridge/CMakeLists.txt#L29

It should be `8.4.0`.

mjcarroll:
> do we need to set a min required ign-msgs8 version here?

Bumped minimum version here: e0f648e

ros_ign_bridge/include/ros_ign_bridge/convert/ros_ign_interfaces.hpp
```diff
@@ -19,13 +19,15 @@
...
 #include <ros_ign_interfaces/msg/contacts.hpp>
+#include <ros_ign_interfaces/msg/dataframe.hpp>
 #include <ros_ign_interfaces/msg/light.hpp>
 
 // Ignition messages
 #include <ignition/msgs/entity.pb.h>
 #include <ignition/msgs/joint_wrench.pb.h>
 #include <ignition/msgs/contact.pb.h>
 #include <ignition/msgs/contacts.pb.h>
+#include <ignition/msgs/dataframe.pb.h>
```
chapulina: The `galactic` branch needs to compile against Edifice and Fortress. So we'll need `ifdef`s here since this message isn't in Edifice.

chapulina (changes requested):
> CI may fail without the updated ign-msgs package here

For Edifice, we'll need `ifdef`s to exclude this message.

For Fortress, we'll need to port the latest `ign-msgs8` binaries to packages.ros.org using reprepro.

ros_ign_bridge/include/ros_ign_bridge/convert/ros_ign_interfaces.hpp
mjcarroll: I added a `HAVE_DATAFRAME` variable that is only set when ignition-msgs 8.4.0 or greater is installed in 565a6bb

This should allow edifice and fortress to build so that we can go ahead and merge this to galactic.  Once ign-msgs > 8.4.0 gets synced out, this functionality will come along without additional modifications.

mjcarroll:
`Gpr` is red because of the OpenGL warning that hasn't been addressed yet.

iche033 (approved):
latest changes look good to me.

ros_ign_bridge/include/ros_ign_bridge/convert/ros_ign_interfaces.hpp
chapulina: That works for me, thanks!

PR #240, opened 2022-04-12
Adding ign_gazebo_dev package

### LINKED ISSUE #186: Provide a package that pulls the default Ignition version for each ROS distro
Each ROS distro is paired with a specific Ignition version as described on [REP-2000](https://www.ros.org/reps/rep-2000.html). It's the same for Gazebo Classic.

We make sure that `ros_ign` releases match the official version. But since multiple Ignition rosdep keys may be available to packages across ROS distros (because those distros target the same platforms, i.e. Foxy and Galactic are on Ubuntu Focal), there's the potential for users mixing up versions unintentionally.

In order to make downstream maintainers' lives easier, we could provide a package whose sole purpose is to pull the correct Ignition version for a given ROS distro. Another advantage is that as long as the underlying APIs don't change, users can seamlessly upgrade from one ROS distro to another without editing their packages.

## Desired behavior

Instead of depending directly on a Gazebo version, i.e.

```
<depend>ignition-gazebo5</depend>
```

Depend on the wrapper:

```
<depend>ros_ign_dev</depend>
```

## Alternatives considered

* Keep the status quo: depending on specific versions of the simulator
* Instead of offering a single package that installs all of Ignition, we could have one per library, i.e.
    * `ros_ign_math`
    * `ros_ign_gazebo`
    * `ros_ign_rendering`
    * ...

## Implementation suggestion

This suggestion is inspired by the `gazebo_dev` package:

 https://github.com/ros-simulation/gazebo_ros_pkgs/tree/noetic-devel/gazebo_dev

Here's some history on it:

* https://github.com/ros-simulation/gazebo_ros_pkgs/pull/473
* https://github.com/ros-simulation/gazebo_ros_pkgs/pull/571

harshmahesheka:
# 🎉 New feature

Closes #186 

## Summary
First,package.xml has  `<depend condition>` which will make the package depend on the required ign-gazebo version based on Ros distribution and other packages can depend on it.Secondly, ignition_gazebo_dev-extras.cmake under cmake find and set required ign gazebo package so, other packages can simply go  `find_package(ign_gazebo_dev REQUIRED) 
` instead of finding ign-gazebo which will make their packages furthermore compatible with different ROS distros. Currently, I am creating a draft pr for reviews from the community, after that, we can add similar packages for all other ignition libraries.

## Checklist
- [ ] Added tests

## Test it

1. I have created a [test_package](https://github.com/harshmahesheka/test_package/tree/ros_ign_dev), you can clone it or replace find_package and dependencies in any of your own packages. 
2.  Clone this repository in your workspace and do rosdep check to see dependencies change based on your ros distro.
3. Build and source this repository and then build your test package. You can see it and find different gazebo versions based on your ros distro.

### DIFF
--- a/ros_ign_dev/ign_gazebo_dev/cmake/ignition_gazebo_dev-extras.cmake
+++ b/ros_ign_dev/ign_gazebo_dev/cmake/ignition_gazebo_dev-extras.cmake
@@ -1,0 +1,17 @@
+# This code here calls and sets required igniton gazebo based on ROS Distribution so,packages can directly go find_package(ign_gazebo_dev) so,that 
+# they don't need to change code with different ros distros
+if(DEFINED ENV{ROS_DISTRO})    
+message(STATUS "Detected ROS Distro is $ENV{ROS_DISTRO}")
+
+if("$ENV{ROS_DISTRO}" MATCHES "foxy")
+  find_package(ignition-gazebo3 REQUIRED)
+  set(IGN_GAZEBO_VER ${ignition-gazebo3_VERSION_MAJOR})
+elseif("$ENV{ROS_DISTRO}" MATCHES "galactic")
+  find_package(ignition-gazebo5 REQUIRED)
+  set(IGN_GAZEBO_VER ${ignition-gazebo5_VERSION_MAJOR})
+elseif("$ENV{ROS_DISTRO}" MATCHES "humble")
+  find_package(ignition-gazebo5 REQUIRED)
+  set(IGN_GAZEBO_VER ${ignition-gazebo6_VERSION_MAJOR})
+else()
+message(STATUS "No ROS Distro detected")
+endif()
--- a/ros_ign_dev/ign_gazebo_dev/cmake/ignition_gazebo_dev-extras.cmake
+++ b/ros_ign_dev/ign_gazebo_dev/cmake/ignition_gazebo_dev-extras.cmake
@@ -1,7 +1,10 @@
 # This code here calls and sets required igniton gazebo based on ROS Distribution so,packages can directly go find_package(ign_gazebo_dev) so,that 
 # they don't need to change code with different ros distros
-if(DEFINED ENV{ROS_DISTRO})    
+set( ENV{OVERRIDE_ROS_GZ_VERSION} 6 )
+
+if(DEFINED ENV{ROS_DISTRO})
 message(STATUS "Detected ROS Distro is $ENV{ROS_DISTRO}")
+endif()
 
 if("$ENV{ROS_DISTRO}" MATCHES "foxy")
   find_package(ignition-gazebo3 REQUIRED)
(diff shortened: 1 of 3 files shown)


ros_ign_dev/ign_gazebo_dev/cmake/ignition_gazebo_dev-extras.cmake
```diff
@@ -0,0 +1,17 @@
+# This code here calls and sets required igniton gazebo based on ROS Distribution so,packages can directly go find_package(ign_gazebo_dev) so,that 
+# they don't need to change code with different ros distros
+if(DEFINED ENV{ROS_DISTRO})
```
nuclearsandwich: In order to use the `ROS_DISTRO` environment variable in CMake configurations this package should have a `build_depend` on the `ros_environment` package which provides hooks to set `ROS_DISTRO`, `ROS_VERSION`, and `ROS_PYTHON_VERSION`.

ros_ign_dev/ign_gazebo_dev/cmake/ignition_gazebo_dev-extras.cmake
```diff
@@ -0,0 +1,17 @@
...
+if("$ENV{ROS_DISTRO}" MATCHES "foxy")
+  find_package(ignition-gazebo3 REQUIRED)
+  set(IGN_GAZEBO_VER ${ignition-gazebo3_VERSION_MAJOR})
+elseif("$ENV{ROS_DISTRO}" MATCHES "galactic")
+  find_package(ignition-gazebo5 REQUIRED)
+  set(IGN_GAZEBO_VER ${ignition-gazebo5_VERSION_MAJOR})
+elseif("$ENV{ROS_DISTRO}" MATCHES "humble")
+  find_package(ignition-gazebo5 REQUIRED)
+  set(IGN_GAZEBO_VER ${ignition-gazebo6_VERSION_MAJOR})
+else()
```
nuclearsandwich, line 18: It is probably worth allowing an override / fallback for unofficial ROS distributions.

Using a CMake variable like `OVERRIDE_ROS_GZ_VERSION` or similar for override or checking `ROS_GZ_VERSION` environment variable to fall back to if no known ROS_DISTRO is found.

nuclearsandwich (commented):
I left some initial feedback and suggestions. I think this is in good shape!

ros_ign_dev/ign_gazebo_dev/cmake/ignition_gazebo_dev-extras.cmake
harshmahesheka: I have added ros_environment dependency.

ros_ign_dev/ign_gazebo_dev/cmake/ignition_gazebo_dev-extras.cmake
harshmahesheka: I have added ign fortress as an override version in ignition_gazebo_dev-extras.cmake and package.xml.

methylDragon:
Heya @harshmahesheka I noticed this was still in draft mode :o, do we want to move forward with this?

harshmahesheka:
> Heya @harshmahesheka I noticed this was still in draft mode :o, do we want to move forward with this?

Definitely. I was waiting for approval on whether the current iteration is correct. Also, this one is just for ign-gazebo( gz-sim now), so should we create a similar folder for all other libraries in gazebo?

azeey:
Closing this since this has been superceeded by vendor packages.

PR #241, opened 2022-04-12
[galactic] Backport GuiCamera, StringVec, TrackVisual, VideoRecord

mjcarroll:
All of these messages are available in `galactic`.


iche033 (approved):
looks good to me

PR #257, opened 2022-06-07
Add optical frame publisher from subt/mbzirc

mjcarroll:
# 🎉 Add Optical Frame Publisher

## Summary

This is a utility that is used frequently enough, I think that it makes sense to include here.  This transforms images (and optionally camera info) into the ROS coordinate conventions from those that gazebo uses.

I have written it as a component, which can take advantage of intraprocess communication where available.

## Checklist
- [ ] Added tests


ros_ign_utils/package.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>ros_ign_utils</name>
```
chapulina: Let's not create anything new with `ign` on the name

ros_ign_utils/package.xml
```diff
@@ -0,0 +1,27 @@
...
+  <version>0.0.0</version>
+  <description>Utility functions and applications for Gazebo ROS interactions</description>
```
chapulina: This sounds very similar to `ros_ign_gazebo` - what do you imagine the difference between them being?

chapulina (changes requested):

ros_ign_utils/package.xml
mjcarroll: I was under the assumption that `ros_ign_gazebo` was specifically for the tasks of starting the simulator as well as spawning. 

I wouldn't be opposed to moving this under there, but with the note that it will inflate the dependencies of `ros_ign_gazebo` without necessarily being something that every user would need/want.

chapulina: > inflate the dependencies of ros_ign_gazebo

Are you referring to `rclcpp_components`?

> without necessarily being something that every user would need/want.

Yeah it's good to not pay for what you don't use, but if one thing, it probably should be the other way around - the "utils" package has as few dependencies as possible.

chapulina: > ros_ign_gazebo was specifically for the tasks of starting the simulator as well as spawning.

I've been thinking of it more broadly as a package to hold simulation-specific functionality, to separate it from `ros_ign_bridge`, which is only concerned with bridging transport

mjcarroll: > Are you referring to rclcpp_components?

As well as geometry_msgs, sensor_msgs, tf2 and tf2_ros, all of which need to be added to package.xml :upside_down_face:

mjcarroll: > I've been thinking of it more broadly as a package to hold simulation-specific functionality, to separate it from ros_ign_bridge, which is only concerned with bridging transport

In that case, I will move it over, if it's ever an issue, it should be trivial to separate out as necessary.

mjcarroll: Moved in 67e340c

ros_ign_gazebo/src/optical_frame_publisher.cpp
```diff
@@ -0,0 +1,200 @@
...
+  void CheckSubscribers();
+
+  /// \brief Start upstream subscription for images
+  void ImageConnect();
+
+  /// \brief Start upstream subscription for camera info
+  void CameraInfoConnect();
+
+  /// \brief Image callback
+  void UpdateImageFrame(const sensor_msgs::msg::Image & msg);
```
caguero, line 53: Document `msg`.

ros_ign_gazebo/src/optical_frame_publisher.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+  /// \brief Camera info callback
+  void UpdateCameraInfoFrame(const sensor_msgs::msg::CameraInfo & msg);
```
caguero, line 57: Document `msg`.

ros_ign_gazebo/src/optical_frame_publisher.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+  /// \brief Publish new TF between frame and optical_frame
+  void PublishTF(const std::string & frame, const std::string & child_frame);
```
caguero, line 62: Document `frame` and `child_frame`.

ros_ign_gazebo/include/ros_ign_gazebo/optical_frame_publisher.hpp
```diff
@@ -0,0 +1,46 @@
...
+namespace ros_ign_gazebo
+{
+
+/// \brief ROS Node to transform image coordinate conventions to match
+/// ROS REP 103 (https://www.ros.org/reps/rep-0103.html)
+class OpticalFramePublisher : public rclcpp::Node
+{
+public:
+  /// \brief Constructor
+  explicit OpticalFramePublisher(const rclcpp::NodeOptions & options);
```
caguero, line 35: Document `options`.

ros_ign_gazebo/src/optical_frame_publisher.cpp
```diff
@@ -0,0 +1,200 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+#include "ros_ign_gazebo/optical_frame_publisher.hpp"
+
+#include <tf2/LinearMath/Quaternion.h>
+#include <tf2_ros/transform_broadcaster.h>
```
caguero: Do you actually need this `include`?

ros_ign_gazebo/src/optical_frame_publisher.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+  if (info_pub && info_pub->get_subscription_count() > 0u) {
+    CameraInfoConnect();
+  }
+}
+
+void OpticalFramePublisher::Impl::ImageConnect()
+{
+  // Skip connecting if publisher is already created (non-nullptr)
+  if (image_pub) {
```
caguero: Shouldn't this be `image_sub`? I think we should skip if we already created the subscriber, right?

ros_ign_gazebo/src/optical_frame_publisher.cpp
```diff
@@ -0,0 +1,200 @@
...
+  image_sub = rclcpp::create_subscription < sensor_msgs::msg::Image > (
+    node_topics,
+    "input/image", 10,
+    std::bind(&OpticalFramePublisher::Impl::UpdateImageFrame, this, _1));
+}
+
+void OpticalFramePublisher::Impl::CameraInfoConnect()
+{
+  // Skip connecting if publisher is already created (non-nullptr)
+  if (info_pub) {
```
caguero: Shouldn't this be `info_sub`? I think we should skip if we already created the subscriber, right?

ros_ign_gazebo/src/optical_frame_publisher.cpp
```diff
@@ -0,0 +1,200 @@
...
+  void PublishTF(const std::string & frame, const std::string & child_frame);
+
+  /// \brief Interface for creating publications/subscriptions
+  rclcpp::node_interfaces::NodeTopicsInterface::SharedPtr node_topics;
+
+  /// \brief timer with callback to check for publisher subscription count
+  rclcpp::TimerBase::SharedPtr timer;
+
+  /// \brief Static transform broadcaster that broadcasts the optical frame id
+  std::unique_ptr<tf2_ros::StaticTransformBroadcaster> tfBroadcasterStatic;
```
caguero: Should we unify the naming style of the member variables (camel or snake)?

ros_ign_gazebo/src/optical_frame_publisher.cpp
mjcarroll: No, removed.

ros_ign_gazebo/src/optical_frame_publisher.cpp
mjcarroll: Yes, updated.

ros_ign_gazebo/src/optical_frame_publisher.cpp
mjcarroll: Changed to snake case, since that's what it mostly was

caguero (commented):
:red_circle:  CI is unhappy :red_circle:

mjcarroll:
This is going to be retargeted to galactic, and I have also made an effort to robustify tests here: https://github.com/gazebosim/ros_gz/pull/272

ahcorde:
what's the state of this PR @mjcarroll ?

mjcarroll:
It just fell off my radar, probably just needs one more round of review and it's good to go.

caguero (commented):
It looks good to me. Should we consider a test?

ahcorde (changes requested):
@mjcarroll This PR is targeting `galactic` which is EOL. Can we close this PR or can you retarget this to `ros2`?

ahcorde:
This PR is targeting `galactic` which is EOL, feel free to open a PR against `rolling` if this PR is still relevant

Issue #276, opened 2022-07-22
Missing dependencies for ros_ign_gazebo_demos


rsinnet:
## Environment
* OS Version: All
* Source or binary build? All

## Description
* Expected behavior: rosdep install should install all necessary packages to run the demos
* Actual behavior: rosdep install completes but some unlisted dependencies are missing

## Steps to reproduce
1. Install minimal version of ROS.
2. Clone this repo into a new catkin workspace
3. Run rosdep install
4. Run roslaunch on pretty much any demo

chapulina:
* https://github.com/gazebosim/ros_gz/pull/277

PR #277, opened 2022-07-22
Fix catkin dependencies for demos

rsinnet:
Add missing dependency on rqt_topic for ros_ign_gazebo_demos to package.xml.
# 🦟 Bug fix

Fixes: #276

## Summary
Added missing packages to package.xml for demos package.

## Checklist
- [x] Added tests
- [x] Updated migration guide (as needed)
- [x] Consider updating Python bindings (if the library has them)

### DIFF
--- a/ros_ign_gazebo_demos/package.xml
+++ b/ros_ign_gazebo_demos/package.xml
@@ -23,6 +23,7 @@
   <!--exec_depend>ros_ign_point_cloud</exec_depend-->
   <exec_depend>rqt_image_view</exec_depend>
   <exec_depend>rqt_plot</exec_depend>
+  <exec_depend>rqt_topic</exec_depend>
   <exec_depend>rviz</exec_depend>
 
   <replace>ros1_ign_gazebo_demos</replace>


ahcorde (changes requested):
I don't think we need to add some of the dependencies. For example `*_controllers` is something that is used with `gz_ros2_control` but these examples are not using this library.

Can you point me where is this package using all these controllers  ?

rsinnet:
> I don't think we need to add some of the dependencies. For example `*_controllers` is something that is used with `gz_ros2_control` but these examples are not using this library.
> 
> Can you point me where is this package using all these controllers ?

Oops, sorry! I think I mixed this up with: https://github.com/gazebosim/ros_gz/pull/277

I updated this current PR with the single missing dependency on rqt_topic.

ahcorde (approved):

rsinnet:
> > I don't think we need to add some of the dependencies. For example `*_controllers` is something that is used with `gz_ros2_control` but these examples are not using this library.
> > Can you point me where is this package using all these controllers ?
> 
> Oops, sorry! I think I mixed this up with: #277
> 
> I updated this current PR with the single missing dependency on rqt_topic.

I see I put the wrong issue link. For future reference, this is the other one I was working on simultaneously: https://github.com/ros-controls/ign_ros_control/pull/2

PR #262, opened 2022-06-22
ign -> gz : ros_gz Migration

methylDragon:
# REBASE MERGE THIS!! Files have moved!!

Part of: https://github.com/gazebo-tooling/release-tools/issues/698

This is a two-parter:
1. Migrate `ros_ign` -> `ros_gz`
2. Leave behind a shim `ros_ign` that redirects the following to `ros_gz`:
    - Launchfiles
    - Executables
    - Libraries?
    - CMake variables

Ideally, the shim should be resolvable via rosdep, CMake, and apt, and pull in `ros_gz` functionality.

# Split PRs
- https://github.com/gazebosim/ros_gz/pull/279

# Targeted PRs
- [x] https://github.com/gazebosim/ros_gz/pull/279
- [x] https://github.com/gazebosim/ros_gz/pull/280
- [x] https://github.com/gazebosim/ros_gz/pull/282
- [x] https://github.com/gazebosim/ros_gz/pull/281

# Notes
- To make things easier to backport, this does NOT migrate the headers pulled in (`ignition/XXX`), or the namespaces (`ignition::`), or anything relating to the core Gazebo (`gz`) libraries.

:warning: Rebase-merge this!! There are files moved!
When it's time to, and everything is ok, I'll probably squash everything after the move commit to a single commit so we don't dirty the main commit history.

PR #280, opened 2022-07-25
ign -> gz : ros_gz Migration (Local Source)

methylDragon:
Part of: #262

Migrates:
- Namespaces
- Class names and variables

Also:
- Tick-tocks msg type matching

### DIFF
--- a/ros_ign_bridge/src/service_factories/ros_ign_interfaces.cpp
+++ b/ros_ign_bridge/src/service_factories/ros_ign_interfaces.cpp
@@ -45,21 +45,6 @@ get_service_factory__ros_ign_interfaces(
     >(ros_type_name, "ignition.msgs.WorldControl", "ignition.msgs.Boolean");
   }
 
-  // NOTE(CH3): Prioritising ignition since versions prior to Garden cannot support gz
-
-  if (
-    ros_type_name == "ros_ign_interfaces/srv/ControlWorld" &&
-    (gz_req_type_name.empty() || gz_req_type_name == "gz.msgs.WorldControl") &&
-    (gz_rep_type_name.empty() || gz_rep_type_name == "gz.msgs.Boolean"))
-  {
-    return std::make_shared<
-      ServiceFactory<
-        ros_ign_interfaces::srv::ControlWorld,
-        ignition::msgs::WorldControl,
-        ignition::msgs::Boolean>
-    >(ros_type_name, "gz.msgs.WorldControl", "gz.msgs.Boolean");
-  }
-
   return nullptr;
 }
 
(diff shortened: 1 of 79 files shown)


methylDragon:
There are quite a few uncrustify lint errors that don't originate from the migration, if we want to deal with them we should merge everything else in first (otherwise we'll get a whole bunch of conflicts)

There's also a couple of include order lint errors from before the migration that I didn't touch.

ros_ign_point_cloud/src/point_cloud.cc
```diff
@@ -33,13 +33,13 @@
 #include <sensor_msgs/fill_image.h>
 #include <sensor_msgs/point_cloud2_iterator.h>
 
-IGNITION_ADD_PLUGIN(
-    ros_ign_point_cloud::PointCloud,
+GZ_ADD_PLUGIN(
```
chapulina: This breaks Fortress support - you won't notice on CI because this package isn't being built

ros_ign_bridge/src/service_factories/ros_ign_interfaces.cpp
```diff
@@ -44,25 +44,41 @@ get_service_factory__ros_ign_interfaces(
         ignition::msgs::Boolean>
     >(ros_type_name, "ignition.msgs.WorldControl", "ignition.msgs.Boolean");
   }
+
+  // NOTE(CH3): Prioritising ignition since versions prior to Garden cannot support gz
```
chapulina: How about we keep using `ignition` for now, which will result in warnings in Garden but hopefully goes through? And when we backport whatever we can to Citadel and Fortress, we flip the switch here and use `gz` for all versions.

chapulina (changes requested):

ros_ign_point_cloud/src/point_cloud.cc
methylDragon: Ah.. Reverting

ros_ign_bridge/src/service_factories/ros_ign_interfaces.cpp
methylDragon: Okay, note that the `ignition` space will only work if the `include/ignition` version of the message header is used, which... should be fine I think?

methylDragon: Ah, I didn't notice the message generation workflow changed. Hopefully it works: https://github.com/gazebosim/ros_gz/pull/280/commits/1451eb70bc7bd8d8f9cb02eef54ba93fdf2d57f7

ros_ign_point_cloud/src/point_cloud.cc
methylDragon: https://github.com/gazebosim/ros_gz/pull/280/commits/1451eb70bc7bd8d8f9cb02eef54ba93fdf2d57f7

chapulina (approved):
LGTM with happy CI

PR #293, opened 2022-09-08
Update melodic readme to only reflect ROS 1

mjcarroll:
Small documentation tweak.  I think after this, we should also make `ros2` the default branch.


ahcorde (approved):

mjcarroll:
CI is in a bad state, but I don't believe that it has anything to do with updating READMEs

PR #303, opened 2022-09-27
replace ign with gz in ros_gz_bridge README

oKermorgant:
`ros_gz_bridge` documentation in `ros2` and `humble` branches still use the IGN prefix for the `config_file` that may be passed to `parameter_bridge`.


mjcarroll (approved):
Good catch, thanks!

PR #321, opened 2022-10-31
Fix stopwatch exports

mjcarroll:
Add missing exports for the stopwatch library.


mjcarroll:
This actually needs to be forward-ported from galactic.

PR #375, opened 2023-03-15
populate imu covariances when converting

ejalaa12:
# 🎉 New feature

## Summary

This PR depends on those two other PR:
- https://github.com/gazebosim/gz-sensors/pull/333
- https://github.com/gazebosim/gz-msgs/pull/333

Once those are merged, this PR will allow forwarding the IMU covariance between ros <--> gz.

## Checklist
- [x] Added tests
- [x] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))

### DIFF
--- a/ros_gz_bridge/src/convert/sensor_msgs.cpp
+++ b/ros_gz_bridge/src/convert/sensor_msgs.cpp
@@ -274,6 +274,16 @@ convert_ros_to_gz(
   convert_ros_to_gz(ros_msg.orientation, (*gz_msg.mutable_orientation()));
   convert_ros_to_gz(ros_msg.angular_velocity, (*gz_msg.mutable_angular_velocity()));
   convert_ros_to_gz(ros_msg.linear_acceleration, (*gz_msg.mutable_linear_acceleration()));
+
+  for (const auto & elem : ros_msg.linear_acceleration_covariance){
+    gz_msg.mutable_linear_acceleration_covariance()->add_data(elem);
+  }
+  for (const auto & elem : ros_msg.orientation_covariance){
+    gz_msg.mutable_orientation_covariance()->add_data(elem);
+  }
+  for (const auto & elem : ros_msg.angular_velocity_covariance){
+    gz_msg.mutable_angular_velocity_covariance()->add_data(elem);
+  }
 }
 
 template<>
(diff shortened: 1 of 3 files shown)


ros_gz_bridge/test/utils/ros_test_msg.cpp
```diff
@@ -956,6 +959,12 @@ void compareTestMsg(const std::shared_ptr<sensor_msgs::msg::Imu> & _msg)
   compareTestMsg(_msg->orientation);
   compareTestMsg(std::make_shared<geometry_msgs::msg::Vector3>(_msg->angular_velocity));
   compareTestMsg(std::make_shared<geometry_msgs::msg::Vector3>(_msg->linear_acceleration));
+
+  for (int i = 1; i < 10; ++i){
```
ahcorde: why not 0 to 9 ?

ros_gz_bridge/src/convert/sensor_msgs.cpp
```diff
@@ -288,6 +298,30 @@ convert_gz_to_ros(
   convert_gz_to_ros(gz_msg.linear_acceleration(), ros_msg.linear_acceleration);
 
   // Covariances not supported in gz::msgs::IMU
+  int data_size = gz_msg.linear_acceleration_covariance().data_size();
+  if (data_size == 9)
```
ahcorde: can you define 9 as a const?

ahcorde (changes requested):

ros_gz_bridge/test/utils/ros_test_msg.cpp
ejalaa12: Nice catch, it was because in the `createTestMsg` method above, the expected covariance arrays are {1,2,3....,9}.
But I can instead loop from 0 to 9 and do 
```cpp
EXPECT_EQ(_msg->...covariance[i] = i+1);
...
```
By the way, it is weird that the unit test passed. I'll double check

ros_gz_bridge/src/convert/sensor_msgs.cpp
ejalaa12: again why a const instead of a simple literal? IMO, introducing a variable is not worth it in this context.
What do you think ? @ahcorde

ros_gz_bridge/test/utils/ros_test_msg.cpp
ejalaa12: So I double checked, the unit test were indeed not passing. I fixed them, as explained above.
But other tests were failing as well ( NavSat & GpsFix), so I went ahead and fixed them as well.

ejalaa12:
Again, note that the CI will not pass until the other 2 PRs are merged.

mjcarroll:
Getting CI set up for harmonic here.  We are going to need to make sure that we #ifdef the appropriate bits because this message is only supported in harmonic+

ejalaa12:
Do you have an example of this somewhere in the code ? What is the token I should check for ? I couldn't find one.

mjcarroll:
> Do you have an example of this somewhere in the code ? What is the token I should check for ? I couldn't find one.

I just pushed an implementation to your branch.  This isn't something we've really done much, so there isn't documentation on it.

I added a define based on the MAJOR version number of gz-msgs, as covariances are only available in gz-msgs10+

nkoenig (approved):
Looks good

mjcarroll:
One more test failure here, @ejalaa12 can you take a look?

I think this repo isn't quite as subject to the freeze as there will be no harmonic/rolling binaries generated from here.

ejalaa12:
@mjcarroll, sorry for the long delay. I just pushed the fix, all checks are passing now :)

mjcarroll (approved):

ahcorde (approved):

PR #386, opened 2023-05-04
Port: humble to ros2

mjcarroll:


ros_gz/CHANGELOG.rst
```diff
@@ -9,6 +9,12 @@ Changelog for package ros_gz
 * Merge remote-tracking branch 'origin/humble' into ahcorde/humble_to_ros2
 * Contributors: Alejandro Hernández Cordero, ahcorde
 
+0.244.10 (2023-05-03)
```
caguero, line 12: Do these empty blocks make sense? Same in other changelogs

caguero (dismissed):
It looks good to me besides a minor comment about the changelogs.

caguero:
Is CI supposed to fail like this?

ros_gz/CHANGELOG.rst
mjcarroll: Yes, this is generally how it is done with ROS releases.  All packages in a repository have to be released at the same time, even if not all the packages have changes.  This is indicated with a blank changelog entry.

caguero (approved):

PR #388, opened 2023-05-05
Fix humble branch names in humble branch

mjcarroll:


moriarty:
I think this PR needs to be retargetted from `ros2` to `humble` based on the title, otherwise this is duplicate of #387

moriarty:
Interesting, looks like branch protection isn't setup on `humble`
```
Review requested
Review has been requested on this pull request. It is not required to merge
```

PR #391, opened 2023-05-19
Add missing rosidl_cmake dep to ros_gz_bridge

Yadunund:
Address https://build.ros2.org/job/Ibin_ujv8_uJv8__ros_gz_bridge__ubuntu_jammy_arm64__binary/1/

### DIFF
--- a/ros_gz_bridge/package.xml
+++ b/ros_gz_bridge/package.xml
@@ -14,6 +14,7 @@
 
   <buildtool_depend>ament_cmake</buildtool_depend>
   <buildtool_depend>pkg-config</buildtool_depend>
+  <buildtool_depend>rosidl_pycommon</buildtool_depend>
 
   <depend>actuator_msgs</depend>
   <depend>geometry_msgs</depend>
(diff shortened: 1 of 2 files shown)


ahcorde:
@Yadunund Is this a new package on Iron ? or is this package already released on Humble ?

clalancette (dismissed):
Though this would likely resolve the build warning, this isn't quite right for `iron`.

Instead, what should happen here is that this package should take a dependency on `rosidl_pycommon`, and the code in https://github.com/gazebosim/ros_gz/blob/e84dda62a8b066f3d218920ca70d6623f08c2b05/ros_gz_bridge/ros_gz_bridge/__init__.py#L21 should be updated to be `from rosidl_pycommon import expand_template`.

To answers @ahcorde's question; this deprecation is new in `iron`, so this should only be released into `rolling` and `iron`.

clalancette:
I just pushed 158bd3b, which I think is the proper fix here.

ahcorde:
CI fixed here https://github.com/gazebosim/ros_gz/pull/395

ahcorde (approved):

mjcarroll (approved):

Issue #461, opened 2023-11-14
ros_gz for humble/harmonic 


0mnimon:
As per gazebosim doc it is mentioned that you can download ros_gz for humble/harmonic using "ros-humble-ros-gzharmonic"
I tried both prerelease and released repo but both are providing error that "the command does not exist"
How do I get ros_gz for humble/harmonic?

![image](https://github.com/gazebosim/ros_gz/assets/150802301/5c69c8dc-dab9-4652-9585-937f8f21abf5)

azeey:
@0mnimon, thanks for the issue. It's an error in the documentation. Currently the Humble/Harmonic combination is not supported.

azeey:
Fix: https://github.com/gazebosim/docs/pull/407

0mnimon:
@azeey Thank you for the reply, will the combination be available any time soon?

azeey:
https://github.com/gazebosim/ros_gz/pull/462

quantumxt:
Hi there, just wondering if there's a binary for `arm64` architecture? 

I compared the following Packages list:
- `amd64` (https://packages.osrfoundation.org/gazebo/ubuntu-stable/dists/jammy/main/binary-amd64/) 
- `arm64` (https://packages.osrfoundation.org/gazebo/ubuntu-stable/dists/jammy/main/binary-arm64/)

`ros-humble-ros-gzharmonic` was only listed inside `amd64`, but not `arm64`.

PR #458, opened 2023-10-27
Added messages for 2D Bounding Boxes to ros_gz_bridge

wittenator:
# 🎉 New feature

Closes #457

## Summary
I added the bridging of Detection2D and Detection2DArray to AnnotatedAxisAligned2DBox and AnnotatedAxisAligned2DBox_V and vice versa. 

## Test it
WIP

## Checklist
- [X] Signed all commits for DCO
- [X] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [X] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [x] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸

PR #467, opened 2023-12-05
Forward port humble->iron

azeey:
# ➡️ Forward port

Port humble to iron
Not sure if we do forward ports in this repo, but I thought it was the best way to get the changes from #461, https://github.com/gazebosim/ros_gz/pull/458, and potentially other changes into `iron`.

Question:
* There were changes in `Changelog` files that I removed. Should I include them? Same for version changes in `package.xml` where the merge brought in 0.247. 

Branch comparison: https://github.com/gazebosim/ros_gz/compare/iron...humble

**Note to maintainers**: Remember to **Merge** with commit (not squash-merge or rebase)

🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸


mjcarroll:
> Not sure if we do forward ports in this repo

We do a little of both in this one because gazebo does forward ports and ros does backports.  It can be a bit muddled, but it works.

ros_gz/package.xml
```diff
@@ -4,7 +4,7 @@
   <!-- TODO: Make this a metapackage, see
        https://github.com/ros2/ros2/issues/408 -->
   <name>ros_gz</name>
-  <version>0.247.0</version>
+  <version>0.245.0</version>
```
mjcarroll: I'm not actually sure what we want to do here.  `245` is the latest release in iron and `247` is the latest in humble.

I think we want to keep the newer version and release it.  @ahcorde what do you think

ahcorde: I think we should modify the version to `1.X`, `2.X`, ... sometimes is hard to find a version which is not already released

azeey: According to https://github.com/gazebosim/ros_gz/blob/ros2/RELEASING.md#versioning, the versions for Iron should start with `0.254`. Do we want to keep that convension?

azeey: I've reverted the change.

mjcarroll (approved):

PR #473, opened 2023-12-22
Update README to include harmonic

luca-della-vedova:
This few liner just updates the README to include harmonic.
I admittedly didn't test all the configurations, instead I looked at the [package.xml from ros_gz_bridge](https://github.com/gazebosim/ros_gz/blob/ros2/ros_gz_bridge/package.xml) to try to guess since `humble`, `iron` and `ros2` all seem to have a branch for detecting and building with Harmonic


luca-della-vedova:
I just noticed https://gazebosim.org/docs/harmonic/ros_installation that shows that Humble and Harmonic are not supported, however the `humble` branch package.xml [seems to have it](https://github.com/gazebosim/ros_gz/blob/humble/ros_gz_bridge/package.xml), so not sure what the status is there.

azeey:
That combination was just added recently. We need to update the documentation. Binary packages aren't available yet, but they're coming soon

azeey (approved):

PR #453, opened 2023-09-21
Filter ROS arguments before gflags parsing

mjcarroll:

PR #485, opened 2024-01-08
Release 0.254.0 for `iron`

azeey:
Going to make a release of 0.254.0 off the `iron` branch. I wanted to create a PR before tagging it since I'm bumping the version from 0.247.0 to 0.254.0 to align with the versioning scheme mentioned in https://github.com/gazebosim/ros_gz/blob/70fb5dc86ea87b909523e7411f95506d3553b3cd/RELEASING.md#L36-L80


mjcarroll (approved):

azeey:
hmm, I was going to do the same for `rolling`, but not sure what the version would be. The last release according to the changelogs was 0.246, but looking at the package available on packages.ros.org, it shows 0.245.

mjcarroll:
It looks like everything except the release happened for 0.246?

azeey:
I guess we can set the first digit to 3 to indicate `rolling` and skip the second digit? So it would be `0.34.0`. Alternatively, we can designate 14 as the number for `rolling` going along with the current sequence, but then the version would be 0.2145.0, which would be "older" than 0.245. I guess this whole versioning scheme breaks when the number associated with the ROS or Gazebo distro is no longer a single digit.

azeey:
> It looks like everything except the release happened for 0.246?

Looks like it.

mjcarroll:
Can we just defer doing anything on rolling and release 0.246 there?

We will likely have to do something once 24.04/jazzy is ready, but maybe that's a good time to bump the major

ahcorde (approved):

azeey:
> Can we just defer doing anything on rolling and release 0.246 there?
> 
> We will likely have to do something once 24.04/jazzy is ready, but maybe that's a good time to bump the major

Okay, I will look into releasing 0.246 from the tag. 

I was thinking we needed some of the recent changes to unblock https://github.com/ros-planning/navigation2/pull/3634, but it looks like we don't since #453 was merged after 0.245 and 0.246.

PR #500, opened 2024-02-21
Add a ROS node that runs Gazebo

### LINKED ISSUE #497: ROS node that runs Gazebo using `gz::sim::Server`
## Desired behavior

There are a few problems with using the `gz` CLI interface for starting Gazebo from ROS.

- There is no way to include additional plugins that get loaded at startup. For example, once #490 is complete, it would be nice to automatically load that plugin without requiring users to include that system in their world files
- Some things we want to do require configuring Gazebo before it starts: Examples:
  - If we want to support `package://` URIs in Gazebo, one way of solving it, (1) from https://github.com/ros/sdformat_urdf/pull/20#issuecomment-1696208691, is to add a URI handler to sdformat. But we can't do this without running something before an SDF files are loaded.
  - One idea to make Gazebo Classic to modern Gazebo migration easier is to detect when an attempt is made by Gazebo to load a plugin from gazebo_ros_pkgs or gazebo_plugins and substitute the equivelent plugin from modern Gazebo. One possible way to do this (haven't tested) is to register a handler for the `LoadSdfPlugin` event. However, this needs to be done outside of a system.
- We don't have a way to load an SDF world file as a string. This would be useful if the SDF file was generated from xacro or something similar. This can be easily implemented in a Node that starts Gazebo using `gz::sim::Server`.

## Alternatives considered
- Add extension points in `gz-sim` to allow loading plugins at startup. I think this would be nice to have, but I don't think it'd solve all the problems because somethings need to be done outside of a plugin scope.

## Implementation suggestion

I propose we implement a ROS node that runs Gazebo using `gz::sim::Server`. This would make it a lot easier to customize Gazebo in a way that's optimized for ROS users.

azeey:
# 🎉 New feature

Closes #497

## Summary
- [x] Add gzserver with ability to load an SDF file or string

## TODO
- [ ] Add ability to load additional systems in addition to the ones loaded by default or as part of the SDF world
- [ ] Configure SDFormat to understand `package://` URIs

## Checklist
- [ ] Added tests

### DIFF
--- a/ros_gz_sim/src/gzserver.cpp
+++ b/ros_gz_sim/src/gzserver.cpp
@@ -29,7 +29,6 @@ int main(int _argc, char ** _argv)
   common::Console::SetVerbosity(4);
   sim::ServerConfig server_config;
 
-
   if (!world_sdf_file.empty())
   {
     server_config.SetSdfFile(world_sdf_file);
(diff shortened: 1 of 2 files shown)


ros_gz_sim/src/gzserver.cpp
```diff
@@ -0,0 +1,68 @@
...
+  else if (!world_sdf_string.empty()) {
+    server_config.SetSdfString(world_sdf_string);
+  }
+  else {
+    RCLCPP_ERROR(
+      node->get_logger(), "Must specify either 'world_sdf_file' or 'world_sdf_string'");
+    return -1;
+  }
+
+  // (azeey) Testing if a plugin can be loaded along side the defaults. Not working yet.
```
mjcarroll: Not sure exactly the interaction, but this path may be more in line with what you are trying to do: https://github.com/gazebosim/gz-sim/blob/633ce72171e27e83bf2e0292c9998e036d5da3fc/src/SimulationRunner.cc#L941

azeey:
Since this is now a dependency of #528 and related PRs, I think it's best to merge it as is and work on the other TODO items in follow-up PRs. It would also be nice to add a test, but we are running short on time with the Jazzy release, so I propose we punt on that as well.

caguero (approved):

azeey:
@Mergifyio backport jazzy

PR #559, opened 2024-06-05
Use `ignoreLocalMessages` in the bridge

### LINKED ISSUE #555: GZ->ROS communication doesn't work when using the bridge and gz_server in the same composable container
Part of #544 

When the bridge and gz_server nodes are in the same process, GZ->ROS communication doesn't work because of https://github.com/gazebosim/ros_gz/blob/cfc94bc0dcf880296c37920cf527b881efc7cb4a/ros_gz_bridge/src/factory.hpp#L112-L115

which ignores all `gz-transport` messages published from the same process. This was done to enable bidirectional bridges, so I don't think we can simply remove it.

caguero:
# 🦟 Bug fix

Fixes #555

Requires https://github.com/gazebosim/gz-transport/pull/506

## Summary

Instead of discarding all messages from nodes within the same process as the subscriber, we only discard messages published from the same node (the one used by the bridge).

This way we should be able to use the bridge with composable nodes, where all nodes share the same process with the bridge.

## Checklist
- [ ] Added tests

### DIFF
--- a/ros_gz_bridge/src/factory.hpp
+++ b/ros_gz_bridge/src/factory.hpp
@@ -109,13 +109,13 @@ public:
       const gz::transport::MessageInfo &)> subCb =
       [this, ros_pub](const GZ_T & _msg, const gz::transport::MessageInfo & _info)
       {
-        // Ignore messages that are published from this bridge.
-        if (!_info.IntraProcess()) {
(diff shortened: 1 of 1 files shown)


caguero:
@osrf-jenkins run tests please

azeey (approved):

PR #562, opened 2024-06-19
feat: `override_timestamps_with_wall_time` parameter

### LINKED ISSUE #546: Support for `use_sim_time` = `false`
What are your thought about running the bridge with `use_sim_time` `false`? Would you be open for a PR like this: https://github.com/gazebosim/ros_gz/pull/545

Please give me your thoughts.

Thanks!

reinzor:
# 🎉 New feature

Closes #546 

## Summary
Stamp all outgoing headers with the wall time if parameter `override_timestamps_with_wall_time` is set to `true`.

## Test it
Start the bridge with the additional parameter `override_timestamps_with_wall_time` set to `true`. Note that the outgoing messages have the wall time stamp and not the gazebo time stamp.

## Checklist
- [ ] Added tests

### DIFF
--- a/ros_gz_bridge/src/factory.hpp
+++ b/ros_gz_bridge/src/factory.hpp
@@ -28,6 +28,16 @@
 
 #include "factory_interface.hpp"
 
+template <class T, class = void>
+struct has_header : std::false_type
+{
+};
+
+template <class T>
+struct has_header<T, std::void_t<decltype(T::header)>> : std::true_type
+{
+};
+
 namespace ros_gz_bridge
 {
 
@@ -104,12 +114,13 @@ public:
     std::shared_ptr<gz::transport::Node> node,
     const std::string & topic_name,
     size_t /*queue_size*/,
-    rclcpp::PublisherBase::SharedPtr ros_pub)
+    rclcpp::PublisherBase::SharedPtr ros_pub,
+    bool override_timestamps_with_wall_time)
   {
     std::function<void(const GZ_T &)> subCb =
-      [this, ros_pub](const GZ_T & _msg)
+      [this, ros_pub, override_timestamps_with_wall_time](const GZ_T & _msg)
       {
-        this->gz_callback(_msg, ros_pub);
+        this->gz_callback(_msg, ros_pub, override_timestamps_with_wall_time);
       };
 
     // Ignore messages that are published from this bridge.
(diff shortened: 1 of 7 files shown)


ahcorde (changes requested):
code is not compiling https://github.com/gazebosim/ros_gz/actions/runs/9579762345/job/26418501612?pr=562

azeey:
@Mergifyio update

azeey:
CI issues should hopefully be fixed after https://github.com/ros2/ros2/pull/1574

Timple:
Let's re-run them then :slightly_smiling_face:

azeey:
I just reran it, but I realized the branch is out-of-date. Do you mind merging from `ros2`?

reinzor:
> I just reran it, but I realized the branch is out-of-date. Do you mind merging from `ros2`?

done

azeey:
It seems like there are some CI failures. Do the tests pass locally for you?

reinzor:
> It seems like there are some CI failures. Do the tests pass locally for you?

I am running the tests locally and they seem to fail as well. However, I don't see how this code change would affect the failing tests. Could you help me with this?

azeey:
The error logs indicated that it was failing to create any of the GZ->ROS bridges. Inspecting the exception with a print statement shows that the problem is that we are redeclaring the `override_timestamps_with_wall_time` parameter. My suggestion is to declare the parameter in https://github.com/gazebosim/ros_gz/blob/4c64bbea8041131b36a84692af7064a4abbfddb4/ros_gz_bridge/src/ros_gz_bridge.cpp#L35 and use `get_parameter` to actually retrieve the value. I would also suggest adding a member variable to `BridgeHandle` that stores the value of the parameter and fetch the parameter in the constructor of `BridgeHandle` instead of modifying the function signature.

reinzor:
Thanks for your help @azeey ; I updated the PR with the suggestions you mentioned. Not sure about not changing the function signature though. We need to pass this variable to the factory right?

ros_gz_bridge/src/factory.hpp
```diff
@@ -28,6 +28,16 @@
 
 #include "factory_interface.hpp"
 
+template<class T, class = void>
+struct has_header : std::false_type
```
ahcorde, line 34: include `<type_traits>`

ros_gz_bridge/src/factory.hpp
```diff
@@ -139,10 +150,20 @@ class Factory : public FactoryInterface
...
     const GZ_T & gz_msg,
-    rclcpp::PublisherBase::SharedPtr ros_pub)
+    rclcpp::PublisherBase::SharedPtr ros_pub,
+    bool override_timestamps_with_wall_time)
   {
     ROS_T ros_msg;
     convert_gz_to_ros(gz_msg, ros_msg);
+    if constexpr (has_header<ROS_T>::value) {
+      if (override_timestamps_with_wall_time) {
+        auto now = std::chrono::system_clock::now().time_since_epoch();
```
ahcorde, line 162: include `<chrono>`

ahcorde (changes requested):

ahcorde:
> Thanks for your help @azeey ; I updated the PR with the suggestions you mentioned. Not sure about not changing the function signature though. We need to pass this variable to the factory right?

You are targeting `ros2`, we can change the API's here, but if we need to backport this, we should do it in a different way

reinzor:
> > Thanks for your help @azeey ; I updated the PR with the suggestions you mentioned. Not sure about not changing the function signature though. We need to pass this variable to the factory right?
> 
> You are targeting `ros2`, we can change the API's here, but if we need to backport this, we should do it in a different way

I will leave that up to you as maintainers. Feel free to move the logic to a different place.

ahcorde (approved):

azeey (approved):
LGTM! Thanks for iterating. I realized the API change issue is not a problem since none of the changed headers are considered public API. I think this could actually be backported if needed.

reinzor:
Great thanks! How does this backporting to `jazzy` work; should I just send another PR that cherry-picks [this commit](https://github.com/gazebosim/ros_gz/commit/6daea2caf9f0190f83ec7a73e7ed13307b193029)?

ahcorde:

PR #588, opened 2024-08-13
Wait for create service to be available.

skasperski:
# 🦟 Bug fix

Not sure if this is technically a bug, but the current behavior does not work when used from ros launch files. When create is used in a launch file together with gazebo, and gazebo takes more then 5 seconds to load, the create node will fail.

This change just makes it try again until the gazebo service is available. It will also check if a shutdown is requested and stop waiting in that case.

### DIFF
--- a/ros_gz_sim/src/create.cpp
+++ b/ros_gz_sim/src/create.cpp
@@ -277,4 +277,3 @@ int main(int _argc, char ** _argv)
   RCLCPP_INFO(ros2_node->get_logger(), "Entity creation was interrupted.");
   return 0;
 }
-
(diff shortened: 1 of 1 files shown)


azeey (approved):
Thanks for the contribution! LGTM with green CI

ahcorde (changes requested):
uncrustify is failing https://github.com/gazebosim/ros_gz/actions/runs/10370604025/job/28710628547?pr=588

```
9: --- src/create.cpp
9: +++ src/create.cpp.uncrustify
9: @@ -269 +269 @@
9: -          return 1;
9: +        return 1;
9: @@ -280 +279,0 @@
```

ahcorde (approved):

PR #632, opened 2024-11-04
Refactor air pressure demo

caguero:
Refactors the `air_pressure` demo leveraging the latest improvements in `ros_gz`. If we're happy with it, I'll propagate the refactor to all the demos.

* I replaced the previous launch file from Python to XML because I think it's the recommendation (and looks easier) but I'll still keep some of the demos in Python to offer some variety.
* I added the QoS support as well. I noticed that because of the QoS, rqt_topic isn't able to show any data. This was probably happening before. Does anybody know if we can pass QoS settings to rqt_topic?

## Checklist
- [x] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #663, opened 2024-12-10
Improve argument parsing in Actions

azeey:
# 🦟 Bug fix

## Summary

The `RosGzBridge` and `GzServer` now support different spellings for boolean arguments (`True`, `true`). This also simplifies how conditionals are used to create composable nodes by evaluating the conditionals and using them as regular Python booleans instead of relying on `PythonExpression`. It was actually the `PythonExpression` that was preventing support of boolean arguments spelled `true`/`false`.

This came up in https://github.com/gazebosim/ros_gz/pull/632
The launch file in XML can now use lower case boolean strings:
https://github.com/gazebosim/ros_gz/blob/25b6644b644183e09f517170ce41b5dc6f16d6b8/ros_gz_sim_demos/launch/air_pressure.launch.xml#L1C1-L15C10

The equivalent python launch file for just gzserver and the bridge would be:
```python
from launch import LaunchDescription
from launch_ros.substitutions import FindPackageShare
from ros_gz_sim.actions import GzServer
from ros_gz_bridge.actions import RosGzBridge

def generate_launch_description():

    return LaunchDescription([
        GzServer(world_sdf_file='sensors.sdf', use_composition=True, create_own_container=True),
        RosGzBridge(bridge_name='bridge1', 
                    config_file=[FindPackageShare('ros_gz_sim_demos'), '/config/air_pressure.yaml'],
                    use_composition='true'
                    )
    ])

```

Note the `use_composition` argument can be a python boolean or a string (`True` or `true`). We should probably deprecate the string input at some point since that's not very Pythonic

## Checklist
- [ ] Added tests

### DIFF
--- a/ros_gz_bridge/ros_gz_bridge/actions/ros_gz_bridge.py
+++ b/ros_gz_bridge/ros_gz_bridge/actions/ros_gz_bridge.py
@@ -64,19 +64,26 @@ class RosGzBridge(Action):
         self.__container_name = container_name
         self.__namespace = namespace
 
-        # This is here to allow using strings or booleans as values for boolean variables when the Action is used from Python
-        # i.e., this allows users to do:
+        # This is here to allow using strings or booleans as values for boolean variables when
+        # the Action is used from Python i.e., this allows users to do:
         #   RosGzBridge(bridge_name='bridge1', use_composition='true', create_own_container=True)
-        # Note that use_composition is set to a string while create_own_container is set to a boolean. The reverse would also work.
-        # At some point, we might want to deprecate this and only allow setting booleans since that's
-        # what users would expect when calling this from Python
+        # Note that use_composition is set to a string while create_own_container is set to a
+        # boolean. The reverse would also work.
+        # At some point, we might want to deprecate this and only allow setting booleans since
+        # that's what users would expect when calling this from Python
         if isinstance(create_own_container, str):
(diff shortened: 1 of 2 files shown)


ahcorde (changes requested):
```
10:  - /opt/ros/rolling/bin/ament_flake8 --xunit-file /github/home/ws/build/ros_gz_bridge/test_results/ros_gz_bridge/flake8.xunit.xml
10:         # This is here to allow using strings or booleans as values for boolean variables when the Action is used from Python
10:                                                                                                    ^
10:         # Note that use_composition is set to a string while create_own_container is set to a boolean. The reverse would also work.
10:                                                                                                    ^
10:         # At some point, we might want to deprecate this and only allow setting booleans since that's
10:                                                                                                    ^
10:             self.__create_own_container = normalize_typed_substitution(TextSubstitution(text=create_own_container), bool)
10:                                                                                                    ^
10:             self.__use_composition = normalize_typed_substitution(TextSubstitution(text=use_composition), bool)
10:                                                                                                    ^
10:         kwargs:Dict = super().parse(entity, parser)[1]
10:               ^
10:         create_own_container_eval = perform_typed_substitution(context, self.__create_own_container, bool)
10:                                                                                                    ^
10: 1     E231 missing whitespace after ':'
10: 6     E501 line too long (125 > 99 characters)
```

ahcorde (approved):

ahcorde:

PR #703, opened 2025-03-11
Add ROS parameters for configuring GZ to ROS optical frame conversion

iche033:
**Marked as draft to get feedback**

# 🎉 New feature

## Summary

Adds a number of ROS parameters for configuring GZ to ROS frame conversion.

The main feature is the `publish_optical_frame` ROS parameter, which applies to the GZ to ROS direction. When set to true, the bridge automatically converts the messages to z-forward optical frame. The outgoing messages will have a new frame_id with "_optical" suffix string, and a TF with x-forward to z-forward transform is published using a static transform broadcaster.

Example usage:

```bash
ros2 run ros_gz_bridge parameter_bridge /rgbd_camera/image@sensor_msgs/msg/Image@gz.msgs.Image --ros-args -p publish_optical_frame:=true -r /rgbd_camera/image:=/rgbd_camera/optical/image
```

The bridge sets the `publish_optical_frame` ROS param to `true` and also remaps the original topic to a new name with an `optical` sub-namespace. 

If the user wants to set a different frame id or transform, they can override them using a couple of other ROS parameters:

* `override_frame_id_string` (string) - Overrides frame_id of the outgoing messages
* `override_frame_transform` (double array) - Custom transform [x, y, z, roll, pitch, yaw]

See the `API` section of README.md for more info.

Note that this feature adds 2 new ROS package dependencies:
* tf2
* tf2_ros

## Test it

See the `GZ to ROS Optical frame conversion` section of README.md for more instructions.

## Checklist
- [ ] Added tests


ros_gz_bridge/src/bridge_handle_gz_to_ros.cpp
```diff
@@ -16,9 +16,82 @@
...
+  std::shared_ptr<gz::transport::Node> gz_node,
+  const BridgeConfig & config)
+  : BridgeHandle(ros_node, gz_node, config)
+{
+  ros_node_->declare_parameter<bool>("override_timestamps_with_wall_time", false);
+  ros_node_->declare_parameter<bool>("publish_optical_frame", false);
+  ros_node_->declare_parameter<std::string>("override_frame_id_string", "");
+  // The array of doubles are [x, y, z, roll, pitch, yaw]
+  ros_node_->declare_parameter<std::vector<double>>("override_frame_transform",
+      std::vector<double>{});
```
ahcorde: include `<vector>`

ros_gz_bridge/src/bridge_handle_gz_to_ros.cpp
```diff
@@ -16,9 +16,82 @@
...
 
+BridgeHandleGzToRos::BridgeHandleGzToRos(
+  rclcpp::Node::SharedPtr ros_node,
+  std::shared_ptr<gz::transport::Node> gz_node,
+  const BridgeConfig & config)
+  : BridgeHandle(ros_node, gz_node, config)
+{
+  ros_node_->declare_parameter<bool>("override_timestamps_with_wall_time", false);
+  ros_node_->declare_parameter<bool>("publish_optical_frame", false);
+  ros_node_->declare_parameter<std::string>("override_frame_id_string", "");
```
ahcorde: include `<string>`

ros_gz_bridge/src/bridge_handle_gz_to_ros_parameters.hpp
```diff
@@ -0,0 +1,48 @@
...
+  /// from the original frame. This is done by broadcasting a tf with the
+  /// frame_id set to the frame_id in the original message, and the
+  /// child_frame_id set to the new frame id in the updated outgoing message.
+  /// The new frame id of the outgoing messages must be set via the
+  /// override_frame_id_string parameter.
+  std::optional<geometry_msgs::msg::Transform> override_frame_transform;
+
+  /// \brief Override the header.frame_id field of the outgoing messages with
+  /// this new frame_id string
+  std::string override_frame_id_string;
```
ahcorde: include `<string>`

ahcorde (changes requested):
Code is not compiling:
```bash
    tf2::tf2

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
 ```

File: ros_gz_bridge/src/bridge_handle_gz_to_ros_parameters.hpp (hunk 3 above)
Comment (iche033) [reply ros_gz_bridge/src/bridge_handle_gz_to_ros_parameters.hpp]: 5166feb4704b4cb992bf0d53a5a528829f59de4d

File: ros_gz_bridge/src/bridge_handle_gz_to_ros.cpp (hunk 2 above)
Comment (iche033) [reply ros_gz_bridge/src/bridge_handle_gz_to_ros.cpp]: 5166feb4704b4cb992bf0d53a5a528829f59de4d

File: ros_gz_bridge/src/bridge_handle_gz_to_ros.cpp (hunk 1 above)
Comment (iche033) [reply ros_gz_bridge/src/bridge_handle_gz_to_ros.cpp]: 5166feb4704b4cb992bf0d53a5a528829f59de4d

**Comment by iche033** [comment]
> Code is not compiling:
> 
> ```shell
>     tf2::tf2
> 
>   but the target was not found.  Possible reasons include:
> 
>     * There is a typo in the target name.
>     * A find_package call is missing for an IMPORTED target.
>     * An ALIAS target is missing.
> ```

fixed.

**Review by ahcorde — COMMENTED** [review]
@iche033 is this ready for review?

**Comment by iche033** [comment]
> @iche033 is this ready for review?

Not yet, I'm planning to add more tests. But I am looking for feedback on the proposed ROS parameters for converting to optical frame, mainly If the general direction is ok.

**Comment by peci1** [comment]
I think this is going in good direction!

1. Why is the static TF being re-broadcasted every second? Is there a reason for that (it cannot ever change as it is defined now)?
2. The name of the parameter `publish_optical_frame` is a bit misleading for me (because it not only publishes the frame, but also overwrites it in messages).
3. In the example in readme, isn't the frame (needlessly) published twice - once by the image bridge and once by the camera_info bridge? If the intended way of use includes just setting `override_frame_id_string`, shouldn't one of the bridges use just this instead of `publish_optical_frame`?
4. One thing to note is that this solution pushes the SDF further away from being a complete source of information for a simulation with ROS. With the PR, the SDF would always need the bridges along with it to be usable in ROS. And the configuration of the bridges could not be automatically determined (except blindly enabling `publish_optical_frame` for all image topics). The alternative solution with an extra SDF tag would be more explicit about this. On the other hand, the usage of optical frames is something ROS-specific, so if Gazebo wants to be middleware-agnostic, it is probably okay.

**Comment by iche033** [comment]
> I think this is going in good direction!

Thanks for the feedback!

> 1. Why is the static TF being re-broadcasted every second? Is there a reason for that (it cannot ever change as it is defined now)?

Currently the static tf broadcaster sends a transform and goes out of scope so I think this caused a problem I ran into, which is that the static TF is not latched, and new subscriber couldn't get the TF. I'll look into getting rid of throttling. 

> 2\. The name of the parameter `publish_optical_frame` is a bit misleading for me (because it not only publishes the frame, but also overwrites it in messages).

Maybe `override_and_publish_optical_frame`?

> 3\. In the example in readme, isn't the frame (needlessly) published twice - once by the image bridge and once by the camera_info bridge? If the intended way of use includes just setting `override_frame_id_string`, shouldn't one of the bridges use just this instead of `publish_optical_frame`?

We can remove the need to publish the frame twice if 
1. the `Image` bridge enables `publish_optical_frame`, and sets `override_frame_id_string` to a fixed string value, e.g. `"my_optical_frame"`
2. the `CameraInfo` bridge will then only need to set `override_frame_id_string:="my_optical_frame"`

The problem is that if the user does not set the extra optional `override_frame_id_string` param in the first Image bridge, the optical frame id will be an auto-generated string (done by adding a `_optical` suffix to the original frame_id). This generated string will be difficult to guess by the user. One idea around this is to have a parameter, e.g. `override_frame_id_append_suffix_string`. So something like this will work:

1. the `Image` bridge enables `publish_optical_frame` only
2. the `CameraInfo` bridge set `override_frame_id_append_suffix_string:="_optical"`

> 4\. One thing to note is that this solution pushes the SDF further away from being a complete source of information for a simulation with ROS. With the PR, the SDF would always need the bridges along with it to be usable in ROS. And the configuration of the bridges could not be automatically determined (except blindly enabling `publish_optical_frame` for all image topics). The alternative solution with an extra SDF tag would be more explicit about this. On the other hand, the usage of optical frames is something ROS-specific, so if Gazebo wants to be middleware-agnostic, it is probably okay.

yeah this is a ROS-specific solution to the problem. I think having a way to deal with this in SDF would be nice but it was decided to scope it down and go with this for now.

**Comment by peci1** [comment]
This all seems super complicated. I can imagine lots of frustrated users. Even if they just copy code from tutorials (or ChatGPT does it for them), they forget one thing and the whole camera machinery doesn't work in ROS.

I have an alternative proposition: what if this package provides a convenience launch file that would just get the two Gz and two ROS topic names as arguments and an optional argument with the new frame ID (defaulting to `<topic>_optical_frame`), and this launch file would launch:

1. Image bridge (with `override_frame_id_string`)
2. Camera info bridge (with `override_frame_id_string`)
3. Static TF broadcaster (this would probably have to be a custom node so that it autodetects the GZ frame ID). For efficiency reasons, this should subscribe to the camera info topic and not to the image topic. This node could even subscribe just for one message and the shut down the subscription.
4. Remaps of the GZ topic names to the ROS topic names.

What do you think?

**Comment by iche033** [comment]
I think this does not need to be that complicated for most users. To migrate their existing bridge to use the new optical frame conversion feature, I expect most users to only set the `publish_optical_frame:=true` for both `Image` and `CameraInfo` topics  (and ignore any `override_frame_id_string` params). This will 'magically' work for them. The caveat here is that the `_optical` frame is needlessly published twice like you noted. Not ideal but I think that should be ok?

We can also provide a convenience launch file like the one you proposed. It could be implemented on top of this PR and uses the ROS parameters introduced here.

How does this sound?

**Comment by Danilrivero** [comment]
> I think this does not need to be that complicated for most users. To migrate their existing bridge to use the new optical frame conversion feature, I expect most users to only set the publish_optical_frame:=true for both Image and CameraInfo topics (and ignore any override_frame_id_string params). This will 'magically' work for them. The caveat here is that the _optical frame is needlessly published twice like you noted. Not ideal but I think that should be ok?

Agree on this, most users will care about setting up `publish_optical_frame`, the proposals to override are required though, for flexibility and avoid future fixes directly related to that.

**Comment by azeey** [comment]
@iche033 do we have a path forward on this PR? Not sure what the status is.

**Comment by iche033** [comment]
> @iche033 do we have a path forward on this PR? Not sure what the status is.

I'll try to wrap this up in the next couple of weeks

**Comment by spurnvoj** [comment]
Hi @iche033, do you have any updates on this?

**Comment by iche033** [comment]
I haven't had time to get back to this. Apart from the conflicts, I remember what's left is to update examples or tests to use the new parameters but the core functionality is ready. Still need to find time to work on it

File: ros_gz_bridge/README.md [hunk 4]
```diff
@@ -383,6 +432,45 @@ By changing `chatter` to `/chatter` or `~/chatter` you can obtain different resu
...
+    * type: bool
+    * default: false
+    * direction: GZ to ROS
+    * description: Override the header.stamp field of outgoing messages with
+      wall time.
+* `publish_optical_frame`
+    * type: bool
+    * default: false
+    * direction: GZ to ROS
+    * description: Apply x-forward to z-forward transformation to outgoing
```
david-wb, line 458: @iche033 This all seems fine, but I think the problem can be solved with just the `override_frame_id_string` param, because the transform can easily be added to the urdf. That would make this PR a much smaller change.

david-wb: Also, why is the param called `override_frame_id_string` instead of just `override_frame_id`?

iche033: renamed `override_frame_id_string` to `override_frame_id` [74967fb](https://github.com/gazebosim/ros_gz/pull/703/commits/74967fbcdeda8141ac65c9d6264809b90e6fb710)

I think what I can do is just split the `override_frame_id` part out from this PR and see if we can get that part merged first.

iche033: https://github.com/gazebosim/ros_gz/pull/826

PR #707, opened 2025-03-17
gz_topic_name in cli to param_bridge

aarsht7:
# 🎉 New feature

## Summary

A quick feature that I needed, so here it is, in case it seem useful to the audience,

It adds the possibility of considering `gz_topic_name` through CLI/launch arguments if it differs from the `ros_topic_name` or vice-versa.

The syntax is `/ROS2_TOPIC@ROS_MSG@GZ_MSG@/GZ_TOPIC`, such that ROS2_TOPIC is the name of the ROS 2 topic,  GZ_TOPIC is the Gazebo internal topic, ROS_MSG is the ROS message type for this topic, and GZ_MSG is the Gazebo message type.

It is optional if ROS 2 and GZ topic names are the same and can be neglected.

## Test it

`ros2 run ros_gz_bridge parameter_bridge /ros/chatter@std_msgs/msg/String@gz.msgs.StringMsg@/gz/chatter`
OR
`ros2 run ros_gz_bridge parameter_bridge /ros/chatter@std_msgs/msg/String]gz.msgs.StringMsg@/gz/chatter`
OR
`ros2 run ros_gz_bridge parameter_bridge /ros/chatter@std_msgs/msg/String[gz.msgs.StringMsg@/gz/chatter`

## Checklist
- [ ] Added tests

### DIFF
--- a/ros_gz_bridge/src/parameter_bridge.cpp
+++ b/ros_gz_bridge/src/parameter_bridge.cpp
@@ -28,7 +28,7 @@
 void usage()
 {
   std::cout << "Bridge a collection of ROS2 and Gazebo Transport topics and services.\n\n" <<
-    "  parameter_bridge [<topic@ROS2_type@Ign_type> ..] " <<
+    "  parameter_bridge [<topic@ROS2_type@Ign_type@gz_topic> ..] " <<
     " [<service@ROS2_srv_type[@Ign_req_type@Ign_rep_type]> ..]\n\n" <<
     "Topics: The first @ symbol delimits the topic name from the message types.\n" <<
     "Following the first @ symbol is the ROS message type.\n" <<
@@ -54,6 +54,9 @@ void usage()
     "A bridge from ROS to Gazebo example:\n" <<
     "    parameter_bridge /chatter@std_msgs/String]ignition.msgs" <<
     ".StringMsg\n" <<
+    "A bidirectional bridge example with gz_topic name:\n" <<
+    "    parameter_bridge /ros/chatter@std_msgs/String@gz.msgs@/gz/chatter" <<
+    ".StringMsg\n\n" <<
     "A service bridge:\n" <<
     "    parameter_bridge /world/default/control@ros_gz_interfaces/srv/ControlWorld\n" <<
     "Or equivalently:\n" <<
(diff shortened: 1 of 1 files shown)


ahcorde (changes requested):
hi @aarsht7 

There are several things here:

 - This will break other user, we can't include this feature in Humble
 - If you want to include this you can open a PR againt rolling but
 - you already have this behaviour
```bash
ros2 run parameter_bridge /chatter@std_msgs/msg/String@gz.msgs.StringMsg --ros-args -r /chatter:=<your desired new ros topic>
```  
 
Can you try this last tip and see if this work in your use case.

aarsht7:
Hello @ahcorde ,

I tested it thoroughly, and it does not break the existing behavior on ros humble (at least not on the tests I performed and for my use cases)
I can also create a PR for rolling if it seems like a useful feature.

Here is the scenario of why I needed this.

I am using simulation to dynamically add the bridge for the topics using a launch file to develop a multi-robot simulation that depends on the number of robots and the types of robots. 

I am creating a list of topics to bridge dynamically and passing it to the `parameter_bridge` node as an argument in my launch file. 

For most of the sensors, I can change the topic names in sdf file with the namespace and all, but some of the gz-system plugins do not allow the change of the topic name with the tag, and data is published with some weird gz topic names, e.g.`/model/robot_A_1/battery/robot_A_1/state` (I can not find `<topic>` tag in [class ref of linear battery plugin](https://gazebosim.org/api/sim/9/classgz_1_1sim_1_1systems_1_1LinearBatteryPlugin.html))

So, to make it more readable and easier for the end-user, I am doing the following in my launch file (this is just an example code) using the changes I have made in PR

```Python
robots  = {'A': 2, 'B': 3, 'C': 1}
bridge_topics = [ ]
for key in robots:
    for i in range(1,robots[key]+1):     
        bridge_topics.append(f"/robot_{key}_{i}/battery@sensor_msgs/msg/BatteryState[gz.msgs.BatteryState@/model/robot_{key}_{i}/battery/robot_{key}_{i}/state")

Node(
        package="ros_gz_bridge",
        executable="parameter_bridge",
        arguments=bridge_topics,
        output="screen",
    )

```

azeey:
I agree with @ahcorde. Topic remapping is something that can be done easily with ROS. In your example, you can pass the remapping in the `remappings` argument (see https://docs.ros.org/en/kilted/Tutorials/Intermediate/Launch/Using-ROS2-Launch-For-Large-Projects.html#remapping). It's slightly more verbose, but a well tested and existing solution.

azeey:
Hi @aarsht7, thank you for the contribution, but I'm going to close this PR since there other ways to accomplish this: (1) Use ROS topic remapping, (2) Use the YAML configuration which allows you to set both the ROS and GZ topic names. We want to steer users away form using the command line to specify the bridge configuration, so I don't think we want to be a new feature to it.

PR #681, opened 2025-01-08
Expose header for `GzServer`

roncapat:
Closes #631 

## Summary
Currently, no header file is defined for https://github.com/gazebosim/ros_gz/blob/ros2/ros_gz_sim/src/gzserver.cpp

This means that currently there is no possibility to import the node in a custom main.cpp to be statically composed with other nodes in the same process space, but the user is forced to go with dynamic composition.

This PR exposes the class via its own header file.

CC @mjcarroll @azeey 

## Checklist
- [ ] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [ ] Consider updating Python bindings (if the library has them)
- [ ] `codecheck` passed (See [contributing](https://gazebosim.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://gazebosim.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Agazebosim+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit message to match the pull request summary while retaining `Signed-off-by` messages.

PR #809, opened 2025-11-18
Expose header for `GzServer` (backport #681)

azeey:
Backport of #681 

Needed by #802 

## Summary
Currently, no header file is defined for https://github.com/gazebosim/ros_gz/blob/ros2/ros_gz_sim/src/gzserver.cpp

This means that currently there is no possibility to import the node in a custom main.cpp to be statically composed with other nodes in the same process space, but the user is forced to go with dynamic composition.

This PR exposes the class via its own header file.

**Note to maintainers**: Remember to use **Rebase-and-Merge**.

### DIFF
--- a/ros_gz_sim/src/gzserver.cpp
+++ b/ros_gz_sim/src/gzserver.cpp
@@ -21,62 +21,52 @@
 #include <rclcpp/rclcpp.hpp>
 #include <rclcpp_components/register_node_macro.hpp>
 
-// ROS node that executes a gz-sim Server given a world SDF file or string.
+#include <ros_gz_sim/gzserver.hpp>
+
 namespace ros_gz_sim
 {
-class GzServer : public rclcpp::Node
+
+GzServer::GzServer(const rclcpp::NodeOptions & options)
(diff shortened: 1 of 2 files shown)


ros_gz_sim/src/gzserver.cpp
```diff
@@ -13,70 +13,59 @@
 // limitations under the License.
 
 #include <functional>
-#include <thread>
```
ahcorde, line 16: keep this line

ahcorde (changes requested):
Just a minor comment

ros_gz_sim/src/gzserver.cpp
azeey: Done

ahcorde (approved):

ahcorde:
@azeey do you need a release or a backport ?

azeey:
@ahcorde there are a few more backports I need to make to the jazzy branch, so I'm good for now.

btw, for future reference,  we should do rebase-and-merge instead of squash merge for backport PRs.

PR #872, opened 2026-03-28
Add start_paused option to GzServer

### LINKED ISSUE #800: Consider adding an option to start the simulation paused.
## Desired behavior
Currently looking at the docs for the GzServer launch action, it does not seem possible to start the simulation in a paused state. There may be use cases where this is desired. For instance if I want to record a video of my simulation , I may want to only start playing the simulation after I prepare my screen recorder.

Cnoobmaster69:
## Summary

Adds a new `start_paused` option to `GzServer` so simulations can start in a paused state.

This wires the option through the launch and action layers so it can be used from both Python and XML launch flows, and passed down to `gzserver` execution via `Server::Run(..., paused)`.

Closes #800

## Changes

1. Added a new `start_paused` option in GzServer

* Added ROS parameter handling in `ros_gz_sim/src/gzserver.cpp`.
* Replaced the hardcoded `paused=false` behavior with the new start_paused parameter.

2. Launch/action plumbing (Python + XML)
Exposed and propagated `start_paused` through:
* `ros_gz_sim/launch/gz_server.launch.py`
* `ros_gz_sim/launch/gz_server.launch`
* `ros_gz_sim/launch/ros_gz_sim.launch.py`
* `ros_gz_sim/launch/ros_gz_sim.launch`
* `ros_gz_sim/ros_gz_sim/actions/gzserver.py`

3. Pause-state consistency fix in simulation interface sync
In `ros_gz_sim/src/gz_simulation_interfaces/gazebo_proxy.cpp`:
* update cached world stats from incoming `SerializedStepMap::stats`
* preserve current paused state when sending control/state during canonical-link state synchronization.

This keeps pause behavior consistent during startup/synchronization flows.

## Motivation

Currently, `GzServer` does not expose a straightforward way to start the simulation in a paused state, even though this can be useful for workflows such as:

* preparing a screen recording before starting the simulation
* spawning / configuring entities before unpausing
* headless or scripted startup flows that need deterministic initialization

This PR adds that missing option while keeping the existing default behavior unchanged unless `start_paused` is explicitly enabled.

## Manual validation

Tested on the `jazzy` branch.

Validated the following flow manually:

1. launch `gz_server` with `start_paused:=true`
2. verify the simulation starts paused through:

   * `/stats`
   * `/gzserver/get_simulation_state`
5. unpause through `/world/<world>/control`
6. verify that simulation resumes and behaves normally afterward (included additional entities)

I also verified that the default behavior remains unchanged when `start_paused` is not set.

## Notes

I have not added automated tests yet.
I’d appreciate guidance on the preferred integration / launch test location in ros_gz_sim for this behavior.


ahcorde (changes requested):
can you add a test ?

do you mind target `rolling` ? then we can backport this feature to older distros

Cnoobmaster69:
> can you add a test ?
> 
> do you mind target `rolling` ? then we can backport this feature to older distros

Thanks @ahcorde.

I’ll retarget this to rolling and add a basic test in `ros_gz_sim` before updating the PR.

PR #879, opened 2026-05-06
ci: install ros-base in ubuntu container

Exokern:
The current `ROS2 CI` workflow runs inside an `ubuntu:24.04` container and calls `.github/workflows/build-and-test.sh`.

Right now that script does not install any base ROS packages before running:

`source /opt/ros/$ROS_DISTRO/setup.bash`

…so the job can fail immediately with:

`.github/workflows/build-and-test.sh: line 29: /opt/ros/rolling/setup.bash: No such file or directory`

This PR installs `ros-$ROS_DISTRO-ros-base` before sourcing. That creates `/opt/ros/$ROS_DISTRO` and lets the job proceed to the actual build/test steps.

Repro (current `ros2`):
- `docker run --rm --platform linux/amd64 -t -v "$PWD":/repo -w /repo ubuntu:24.04 bash -lc 'export ROS_DISTRO=rolling; export ROSDEP_ARGS=""; .github/workflows/build-and-test.sh'`

Validation (this change):
- `apt-get install -y ros-rolling-ros-base` succeeds in `ubuntu:24.04` with the repo's configured ROS apt source and provides `/opt/ros/rolling/setup.bash`.

### DIFF
--- a/.github/workflows/build-and-test.sh
+++ b/.github/workflows/build-and-test.sh
@@ -21,6 +21,10 @@ apt-get install -y python3-colcon-common-extensions \
                    python3-rosdep \
                    libcli11-dev
 
+# Ensure the base ROS distro is installed so /opt/ros/$ROS_DISTRO exists
+# before sourcing it below.
+apt-get install -y ros-$ROS_DISTRO-ros-base
+
 rosdep init
 rosdep update
 rosdep install --from-paths ./ -i -y -r --rosdistro $ROS_DISTRO $ROSDEP_ARGS


azeey (commented):
Thanks for the contribution!

I'm not opposed to this change, but I'm curious why the workflow has been working fine till recently. The `rosdep install` below should install everything necessary for the packages being built right? Maybe an implicit dependency was removed somewhere upstream and we need to add it explicitly in our `package.xml`?

azeey:
Rolling has actually moved to Ubuntu 26.04 so we need to update our CI. There are some Gazebo vendor packages that  have failed to build on the ROS buildfarm due to an issue with Ogre (see https://build.ros2.org/job/Rbin_uR64__gz_cmake_vendor__ubuntu_resolute_amd64__binary/). This will require https://github.com/ros/rosdistro/pull/51194.

Exokern:
Thanks — that makes sense.

Given Rolling has moved to Ubuntu 26.04 (and the current vendor buildfarm block is upstream Ogre / ros/rosdistro#51194), this PR’s ubuntu:24.04 + apt install ros-$ROS_DISTRO-ros-base workaround feels like the wrong layer to fix now.

I’m going to close this PR to avoid burning reviewer time. Happy to reopen or send a new PR once the CI base image / Rolling target is clarified and the vendor unblock lands.

Exokern:
Closing as superseded by Rolling->Ubuntu 26.04 CI update + vendor buildfarm unblock.

PR #910, opened 2026-06-27
Dynamic ros gz bridges

### LINKED ISSUE #849: Dynamically create ros_gz bridges
## Desired behavior
Currently, users must manually map every Gazebo topic to a ROS topic via a YAML file or a lengthy `parameter_bridge` launch argument. This is tedious and error-prone.

We want a "lazy" dynamic bridge that automatically mirrors all available Gazebo topics that have a known 1:1 type mapping. The bridge should only activate (perform the heavy lifting) once a subscriber is detected.

## Alternatives considered
Static YAML mapping: The current method, which is too verbose for complex simulations.

## Implementation suggestion

- Implement a node that listens to the Gazebo transport for new topics (both publishers and subscribers).
- Apply "sane default" QoS settings based on message types (e.g., `sensor_msgs` getting `SensorDataQoS`).
- Allow a fallback to the existing manual mapping for complex topics that require custom logic or don't have a 1:1 mapping.

## Additional context
This would essentially allow `ros_gz_bridge` to feel more like the native `gazebo_ros_pkgs` experience where topics "just show up" in ROS.

Tuxliri:
# 🎉 New feature

Closes #849

## Summary

Adds opt-in dynamic discovery of Gazebo Transport topics to `ros_gz_bridge`.

When `create_dynamic_bridges` is enabled, the bridge periodically:

- Discovers Gazebo topics using `TopicList` and `TopicInfo`.
- Maps supported Gazebo message types using the existing generated mappings.
- Creates bridges for topics that appear after startup.
- Prevents duplicate bridge creation.
- Ignores unsupported or ambiguous message types.

Dynamic bridges default to `GZ_TO_ROS`. The direction can be changed using
`dynamic_bridge_direction`, including `BIDIRECTIONAL`.

`parameter_bridge` can now run without positional topic arguments when dynamic
bridging is enabled.

## Usage

```bash
ros2 run ros_gz_bridge parameter_bridge \
  --ros-args \
  -p create_dynamic_bridges:=true
```

Publish from Gazebo:

```bash
gz topic -t /hello_dynamic_bridge \
  -m gz.msgs.StringMsg \
  -p 'data: "hello from Gazebo"'
```

Observe from ROS:

```bash
ros2 topic echo /hello_dynamic_bridge std_msgs/msg/String
```

## Tests

Added launch integration tests covering:

- Discovery of topics created after bridge startup.
- Gazebo-to-ROS message delivery.
- Bidirectional ROS and Gazebo delivery.
- Duplicate bridge prevention.
- Unsupported message types.
- A burst of 50 ordered messages without duplicates.
- Isolation from other ROS and Gazebo processes.

The tests use separate `GZ_PARTITION` and `ROS_DOMAIN_ID` values.

```bash
source /opt/ros/rolling/setup.bash

colcon build --packages-select ros_gz_bridge \
  --cmake-args -DBUILD_TESTING=ON

source install/setup.bash
colcon test --packages-select ros_gz_bridge
colcon test-result --all --verbose
```

Local result:

```text
100% tests passed, 0 tests failed out of 20
```

## Backport Policy

- [x] This **should not** be backported

This is a Jetty roadmap feature introducing new parameters and behavior.

## Checklist

- [x] Screen capture not applicable; this feature has no UI
- [x] Added tests
- [x] Added example and/or tutorial
- [x] Updated documentation
- [x] Migration guide not needed; the feature is opt-in
- [x] Python binding changes are not applicable
- [x] `codecheck` passed
- [x] All tests passed
- [x] Bazel changes are not applicable to this package
- [ ] Reviewed another open Gazebo pull request
- [x] GenAI was used and `Generated-by` trailers are included

**Note to maintainers**: Please use **Squash-Merge** and retain the
`Signed-off-by` and `Generated-by` trailers.

### DIFF
--- a/ros_gz_bridge/src/parameter_bridge.cpp
+++ b/ros_gz_bridge/src/parameter_bridge.cpp
@@ -77,7 +77,9 @@ int main(int argc, char * argv[])
 
   bool create_dynamic_bridges = false;
   bridge_node->get_parameter("create_dynamic_bridges", create_dynamic_bridges);
-  if (filteredArgs.empty() && !create_dynamic_bridges) {
+  std::string config_file;
+  bridge_node->get_parameter("config_file", config_file);
+  if (filteredArgs.empty() && !create_dynamic_bridges && config_file.empty()) {
     usage();
     return -1;
   }
--- a/ros_gz_bridge/src/parameter_bridge.cpp
+++ b/ros_gz_bridge/src/parameter_bridge.cpp
@@ -68,16 +68,19 @@ using BridgeConfig = ros_gz_bridge::BridgeConfig;
 //////////////////////////////////////////////////
 int main(int argc, char * argv[])
 {
-  if (argc < 2) {
-    usage();
-    return -1;
-  }
   // skip the process name in argument processing
   ++argv;
   --argc;
   auto filteredArgs = rclcpp::init_and_remove_ros_arguments(argc, argv);
 
   auto bridge_node = std::make_shared<RosGzBridge>(rclcpp::NodeOptions());
+
+  bool create_dynamic_bridges = false;
+  bridge_node->get_parameter("create_dynamic_bridges", create_dynamic_bridges);
+  if (filteredArgs.empty() && !create_dynamic_bridges) {
+    usage();
+    return -1;
+  }
 
   // Set lazy subscriber on a global basis
   bool lazy_subscription = false;
--- a/ros_gz_bridge/src/ros_gz_bridge.cpp
+++ b/ros_gz_bridge/src/ros_gz_bridge.cpp
@@ -16,16 +16,47 @@
 
 #include <cstddef>
 #include <memory>
+#include <set>
+#include <sstream>
 #include <string>
 #include <vector>
 
 #include "bridge_handle_ros_to_gz.hpp"
 #include "bridge_handle_gz_to_ros.hpp"
+#include "get_mappings.hpp"
 
 #include <rclcpp/expand_topic_or_service_name.hpp>
 
 namespace ros_gz_bridge
 {
+
+namespace
+{
+std::string bridge_key(const BridgeConfig & config)
+{
+  std::stringstream stream;
+  stream << config.ros_topic_name << '|';
+  stream << config.ros_type_name << '|';
+  stream << config.gz_topic_name << '|';
+  stream << config.gz_type_name << '|';
+  stream << static_cast<int>(config.direction);
+  return stream.str();
+}
+
+BridgeDirection parse_bridge_direction(const std::string & direction)
+{
+  if (direction == "BIDIRECTIONAL") {
+    return BridgeDirection::BIDIRECTIONAL;
+  }
+  if (direction == "GZ_TO_ROS") {
+    return BridgeDirection::GZ_TO_ROS;
+  }
+  if (direction == "ROS_TO_GZ") {
+    return BridgeDirection::ROS_TO_GZ;
+  }
+  return BridgeDirection::NONE;
+}
+}  // namespace
 
 RosGzBridge::RosGzBridge(const rclcpp::NodeOptions & options)
 : rclcpp::Node("ros_gz_bridge", options)
@@ -38,6 +69,8 @@ RosGzBridge::RosGzBridge(const rclcpp::NodeOptions & options)
   this->declare_parameter<bool>("expand_gz_topic_names", false);
   this->declare_parameter<bool>("override_timestamps_with_wall_time", false);
   this->declare_parameter<std::string>("override_frame_id", "");
+  this->declare_parameter<bool>("create_dynamic_bridges", false);
+  this->declare_parameter<std::string>("dynamic_bridge_direction", "GZ_TO_ROS");
   this->declare_parameter("bridge_names", std::vector<std::string>());
   const auto names = this->get_parameter("bridge_names").as_string_array();
 
@@ -120,7 +153,7 @@ RosGzBridge::RosGzBridge(const rclcpp::NodeOptions & options)
 
 void RosGzBridge::spin()
 {
-  if (handles_.empty()) {
+  if (!configured_bridges_created_) {
     std::string config_file;
     this->get_parameter("config_file", config_file);
     bool expand_names;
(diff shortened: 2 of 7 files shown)


azeey:
Hi @Tuxliri, thanks for the contribution. However, as mentioned in #849, the issue is actively being worked on by a GSoC student @C88-YQ. To avoid duplication of work, I recommend finding other Github issues to contribute to.

Tuxliri:
Hi @azeey, gotcha. If @C88-YQ hasn’t coded this in yet the PR is pretty much ready for review, I have tested it and is working fine on my side. Might take this task off his plate.

azeey:
@C88-YQ will be reviewing this

ros_gz_bridge/test/test_dynamic_bridge.cpp
```diff
@@ -0,0 +1,297 @@
...
+               rosMessages.end();
+    }));
+
+  ASSERT_TRUE(gzNode.Subscribe<gz::msgs::StringMsg>(
+    topic,
+      [&mutex, &gzMessages](const gz::msgs::StringMsg & _msg)
+      {
+        std::lock_guard<std::mutex> lock(mutex);
+        gzMessages.push_back(_msg.data());
+    }));
```
C88-YQ:
``` suggestion
  std::function<void(const gz::msgs::StringMsg &)> gzCallback =
    [&mutex, &gzMessages](const gz::msgs::StringMsg & _msg)
    {
      std::lock_guard<std::mutex> lock(mutex);
      gzMessages.push_back(_msg.data());
    };
  ASSERT_TRUE(gzNode.Subscribe(topic, gzCallback));
```


I hit a build failure with `gz-transport16` here. The migration notes mention that `Node::Subscribe` changed to a forwarding wrapper in newer gz-transport versions: [Migration.md](https://github.com/gazebosim/gz-transport/blob/11b728b6c7073cde958a7f28cc9952da19fcb697/Migration.md?plain=1#L104-L123).

  Perhaps we can change it like this to make it compatible across different gz-transport versions

ros_gz_bridge/src/parameter_bridge.cpp
```diff
@@ -68,17 +68,20 @@ using BridgeConfig = ros_gz_bridge::BridgeConfig;
...
 
   auto bridge_node = std::make_shared<RosGzBridge>(rclcpp::NodeOptions());
 
+  bool create_dynamic_bridges = false;
+  bridge_node->get_parameter("create_dynamic_bridges", create_dynamic_bridges);
+  if (filteredArgs.empty() && !create_dynamic_bridges) {
+    usage();
+    return -1;
+  }
+
```
C88-YQ, line 86: This change seems to break the existing config-file-only usage:
``` bash
ros2 run ros_gz_bridge parameter_bridge --ros-args -p config_file:=$WORKSPACE/ros_gz/ros_gz_bridge/test/config/full.yaml
```

ros_gz_bridge/src/ros_gz_bridge.cpp
```diff
@@ -234,12 +259,79 @@ void RosGzBridge::spin()
...
+void RosGzBridge::add_dynamic_bridges()
+{
+  std::vector<std::string> gz_topics;
+  gz_node_->TopicList(gz_topics);
+
+  bool lazy;
+  this->get_parameter("lazy", lazy);
+
+  const auto direction_str = this->get_parameter("dynamic_bridge_direction").as_string();
+  const auto direction = parse_bridge_direction(direction_str);
```
C88-YQ, line 285: I wonder if we should consider determining the bridge direction automatically instead. Since Gazebo and ROS topics may have different directions depending on the specific topic, relying on a single `dynamic_bridge_direction` for all topics might not be flexible enough.

ros_gz_bridge/src/ros_gz_bridge.cpp
```diff
@@ -234,12 +259,79 @@ void RosGzBridge::spin()
...
+      continue;
+    }
+
+    std::set<std::string> gz_type_names;
+    for (const auto & publisher : publishers) {
+      const auto gz_type_name = publisher.MsgTypeName();
+      if (!gz_type_name.empty()) {
+        gz_type_names.insert(gz_type_name);
+      }
+    }
```
C88-YQ, line 307: It looks like the current implementation only supports automatically bridging topics published from Gazebo, while topics subscribed by Gazebo are not handled in the same way. This might be another case we need to consider.

ros_gz_bridge/src/ros_gz_bridge.cpp
```diff
@@ -234,12 +259,79 @@ void RosGzBridge::spin()
...
   }
 }
 
+void RosGzBridge::add_dynamic_bridges()
+{
+  std::vector<std::string> gz_topics;
+  gz_node_->TopicList(gz_topics);
+
+  bool lazy;
+  this->get_parameter("lazy", lazy);
```
C88-YQ, line 282: I think the default lazy state for dynamically created bridges should be `true`. Otherwise, it may cause unnecessary overhead, especially when there are many topics.

ros_gz_bridge/src/ros_gz_bridge.cpp
C88-YQ: We could potentially infer the direction based on the publishers and subscribers of each topic. For example, if a topic only has a publisher on the Gazebo side, it could be bridged as GZ_TO_ROS; if it only has subscribers on the Gazebo side, it could be ROS_TO_GZ; and if both exist, we could fall back to BIDIRECTIONAL for now.

ros_gz_bridge/src/ros_gz_bridge.cpp
C88-YQ: We could potentially apply the same logic to the **subscribers** information from `TopicInfo` and get the `gz_type_name` from there as well.

ros_gz_bridge/test/test_dynamic_bridge.cpp
Tuxliri: Good catch, I was mistakenly building against Jetty

azeey: The regression is tracked in https://github.com/gazebosim/gz-transport/issues/880

ros_gz_bridge/src/parameter_bridge.cpp
Tuxliri: Adding a check for empty config file string.

ros_gz_bridge/src/ros_gz_bridge.cpp
Tuxliri: Do you suggest to automatically bridge all compatible ROS topics to Gazebo?

Tuxliri: Or just supporting a mix of GZ_TO_ROS and ROS_TO_GZ based on the gazebo topic direction?

Tuxliri: @C88-YQ ^

